Issue #41, opened 2024-02-22
Service Architecture


josephjclark:
This is a high level issue to discuss where the AI gen server fits in OpenFn's architecture

# Overview

Basically, right, I've identified a number of data generation services which we need to support Lightning and the CLI:

* Template code generator (AI)
* Other AI Generators (TBD)
* Adaptor Docs generator (used by lightning's docs panel)
* Metadata generator (used by lightning's metadata panel)

It seems to me that one server, sitting at `gen.openfn.org` or `data.openfn.org` or something, should provide all these services (or perhaps proxy to them).

# Architecture

This is the architecture we're gonna need:

![image](https://github.com/OpenFn/gen/assets/7052509/f4ce941c-8ada-47b0-a4f5-dc45eb4bd9e9)

Editable link: [https://excalidraw.com/#json=Z43WyHx3CZcjMrqN1wjfK,5Lf2_ZKsWc168Np7C9Fh0Q](https://excalidraw.com/#json=Z43WyHx3CZcjMrqN1wjfK,5Lf2_ZKsWc168Np7C9Fh0Q)

# Questions

###  Is this really all one server?

I think so!

We've talked about making metadata and docs services within lightning. But actually they have nothing to do with lightning.

But we could have a python server handling AI stuff, and worry about the lightning stuff as a separate server later on.

It just seems like we should be able to setup one solution now that we can expand on later.

###  What language should it be in?

Ideally everything would be in one language, probably node (the metadata and adaptor docs services are javascript based, come to think of it)

The existing AI stuff doesn't really need to be python - but the ML community are likely to want to use python to contribute new services.

See https://github.com/OpenFn/gen/issues/39

### Can we set up a really strong JS-python interface?

I think the ideal here is a JS server which has a really strong pattern to call out to python code. So you'd define an endpoint in Javascript, but you'd call out to a python module to generate the result.

Maybe we have a python server running internally, which the JS server proxies to for certain calls.

### What about fine tuning?

The only thing we have so far that I think MUST be python is the fine-tuning we've got set up for the llama and gpt models. The training is all written in python and look relatively heavyweight.

But I also don't think that's part of the webserver API -it's more like an offline script. So it could sit as a python module inside a node repo, for example.

josephjclark:
It might be worth continuing to develop the AI server in python, so that we don't alienate the ML crowd, and deploy a prototype of that. And worry about the other stuff later.

[...]

Issue #34, opened 2024-02-21
Generate (and update?) readme file for new adaptor functions


josephjclark:
At the moment code gen allows us to generate signatures, tests and code for an adaptor.

It would also be helpful to generate or update the adaptor's readme file to take the changes into account . This is particularly useful for new adaptors, where we should be able to generate a stub readme based off a template in the prompt.

Issue #32, opened 2024-02-21
Move important code out of util


josephjclark:
A pattern in the `gen` project at the moment is to over-rely on the `util` folder (and 

This makes it it hard to find important code.

The `util` pattern should apply standalone, re-usable pure functions. IMO you should be able to fully understand a project without looking at anything in utils.

In this repo, lots of key functionality is buried in utils. For example, the prompts which get sent to the models are in `utils`. But prompts is one of the most important parts of the project, one of the things I want to easily discover! They should be pulled out into `prompts.py`. Ditto for models.

To be fair, this issue might just be to move `prompys.py` and `models.py` out of the utils folder :shrug:

PR #44, opened 2024-04-19
New Server

### LINKED ISSUE #42: Create data/AI platform server
This issue is to restructure the `gen` repo so that it contains:

* A web platform written in typescript and bun
* A bunch of standalone python services

## Structure

The structure is bun-first, so the top level has a package json and some scripts

```
gen
├──services
│   ├── codegen
│    │  ├── README.md
│    │  ├── src
│    │  └── main.py
│    └── test
|   ├── ...other python services
├── platform
|   ├── README.md
|   ├── src
|   │   ├── index.ts
|   └── test
| package.json
```

## Web platform

I want a really simple, data-driven layer to hook up web services. This needs experimenting with but it'll be something like:

```
const routes = {
  // endpoint: python module
 'template-gen': loadPython('code_generator')
 'metadata': loadJS('metadata')
}
```

The TypeScript framework will  accept a post request with a JSON payload, do any auth that's appropriate, and feed the JSON payload through to the python service as a dictionary.

This object-dict binding will be the core of the API really. Python modules should be loaded from a predictable folder structure with a main handler function.

At the moment I don't plan to support resources in paths or anything - just static bindings of a path to a python module.

Bun has some frameworks for web servers. Do we need anything fancy or can we just use basic http server APIs? I think the server component is going to be pretty trivial. But it's worth a look around. [Elysia](https://github.com/elysiajs/elysia) and native libraries both look fine.

## Hosting & Branding

This is either a data-generation server which includes AI stuff, or an AI server which includes data stuff. I prefer the former.

But one way around the semantics is to give the server a name, like `apollo`

> God of **oracles**, healing, archery, music and arts, light, **knowledge**, herds and flocks, and protection of the young

Apollo fits nicely into our product pantheon - and the idea of this being a server of prophecy, oracles and knowledge is just perfect.

Athena, as goddess of wisdom, would be be another candidate.

## Authentication

Should we adopt a security-first approach to this server?

Lightning and the CLI are both likely to be the major clients.

But neither will share any particularly sensitive information. Metadata is only as sensitive as the credential you use, adaptor docs should be considered public, and any generated code templates are not sensitive.

We may want Lightning and the CLI to include some kind of token so that we can monitor usage and rates and we have some idea who is calling. CLI's are anonymous of course but maybe each can generate its own identifier.

This excludes any tokens needed for AI generation.

## Data services

The existing python needs to be restructured, removing all the web server stuff and just leaving the core python modules (#30)

A data service can be python (like the current adaptor code generator) or typescript (like the metadata service). 

In either case it is a function which accepts an object with arguments as input and returns a JSON object as output. The returned object may have some conventions.

All services are stateless. They may utilise caching (that probably needs to happen inside the service handler, if appropriate)

## Python

The server will create a python runtime, in which all python code can be implemented.

I'll need to work out the best way to structure things so that poetry can install dependencies from the root.

## CLI integration

I guess this is an issue for kit really, but I want the CLI to integrate well with this.

Something like `openfn apollo <service-name> path/to/json.json -i standard/cli-style/output

Basically it just takes a resource and a payload and it'll ensure the output goes to the right place.

For sevices which write to the adaptors monorepo, we'll either need like a command override (most likely) or there's some special flag you add like `-a salesforce` which causes the results to be written to salesforce.

Maybe the return data from the services can use a convention, like if it returns a bunch of files, they come down like:
```
files: {
   'src/Adaptor.js': '<...>'
}
```

## Checklist

* [ ] Setup a basic bun web server
* [ ] call out to the sig gen python module
* [ ] dockerise
* [ ] Formalize the web APIs
* [ ] Set up a nice declarative router
* [ ] Create one entrypoint for api gen
* [ ] Refactor stuff out of utils
* [ ] remove python server stuff

josephjclark:
This PR creates a bun-based webserver which will host our AI and data-generation services.

Although I've refactored a lot of the python code (stripping out the HTTP layer), the aim of this project is to wrap a new web server around the existing code and not actually change anything. So the actual python services do need work and this is entrirely expeted!

Isma's original work is all committed under https://github.com/OpenFn/gen/releases/tag/last-isma-commit

## Issues

Closes #42 and also #41 #34 #32

## Features

* A simple bun.js/Elysia webserver which automatically routes to python services
* A poetry "monorepo" of scripts, easily invokeable
* Long-running python engine with C bindings from node - so should be efficient
* Hot reloading (in Typescript and Python)
* Call python directly with `bun py inference tmp/path.json` (useful for debug)
* Services to generate an adaptor template
* Load keys from `.env` (designed for dev mode only)
* Docker image

## Index page

If you go to the the server  root, you'll be given a simple index page showing the available services. We show the first paragraph of the readme on the index, and link to the full readme.

image

## Example Usage

All services are hosted at `/services/<service-name>` and accept a POST with JSON:

```bash
curl -X POST localhost:3000/services/adaptor_gen --json @tmp/payload.json
```

With a payload:

```json
{
  "endpoint": "/facts",
  "model": "gpt3_turbo",
  "open_api_spec": "{ .. }"
}
```

This will call into the `adaptor_gen` service and return an adaptor file to you:

```
{
  "Adaptor.js": "..."
}
```

For very easy testing, use the `echo` service!

```bash
curl -X POST localhost:3000/services/echo --json @tmp/payload.json
```

With a payload:

```json
{
  "x": 22
}
```

This will pass your payload into python and return it right back to you.

## Tasks

* [x] Hook up a basic bun server with Elysia
* [x] Bridge from JS to Python 
* [x] Create a docker image
* [x] Automatically generate routes to python modules in services
* [x] Import python modules on-demand (not on startup)
* [x] Always re-import when running in dev mode (this now works in python AND js for seamless hot loading)
* [x] Setup the poetry venv properly and bind node-calls-python to it
* [x] Poetry monorepo setup (is there anything to do here?)
* [x] Fix python module imports (thanks Elias!)
* [x] Setup a very basic dir-style listing at root
* [x] Port Isma's Inference engine
* [x] Add poetry and python to docker image
* [x]  Port Isma's Signature generator
* [x]  Port Isma's Code generator
* [x] Add new adaptor generator service
* [x] Think a bit more about error handling - how are errors caught and reported (test with syntax errors, bad imports, python execptions)
* [x] Maybe support a .env file - I guess up in the root?
* [ ] Work out whether or how I want to control versions (maybe each service has a version, like `services/template-gen/v1/` ? Or maybe the version is in a query param?
* [x] ~~Each service/endpoint will expect json in a certain format. How do we define that, and how do we version it? Does each service need to provide a json schema? Or d.ts? Should we have standard wrappers on the response and request, or can we be super loose?~~ JSON in, JSON out. That's it my friend.
* [x] Basic documentation in readme and key parts of source
* [x] Ensure docs / process / examples for runing python directly? How do we pass data in? What about a built-in CLI that runs a python script but feeds a json object into it? Maybe that's unneccessary
* [x] How-to guide for contributors
* [ ] Ensure good unit test conventions and tooling are setup (I don't think unit testing will be a big deal in this repo)

## Future Work

* [ ] Create a metadata service
* [ ] Create an adaptor docs service
* [ ] Track usage by demanding some kind of API key or unique id from the client

## Issues

Closes #42 (others to follow)


josephjclark:
So I've said "JSON in JSON out" for the endpoints.

But inside the services I'm also saying "what is the point in wrapping everything up in a `{ sngature: }` or ` { code: } ` object?

It's actually annoying internally to have to  wrap everything.

I wonder if there's a common format like:

```
{
   error: { }  // error object of some kind
   result: {}  // the result of whatever process you ran
   files: {} // the result expressed as a number of files, which the CLI will duly create
}
```

That means that each service has a common JSON response format, but has a lot of flexibility within it to do their own thing.

josephjclark:
Cannot begin to tell you how much trouble I've had getting the right versions of bun and python into the docker image :grimacing: 

In the end I've gone with `python:3.11-bullseye`. If anyone feels the need to optimise this later, they can be my guest...

josephjclark:
Outstanding right now is:

python unit tests - but I don't really have anything to test :thinking:  Maybe this can be introduced as needed?

versioning - I think I'll spin this out as an issue, I'm not really sure on the requirements yet. I think service versions are probably tied to server Versions?

josephjclark:
So it looks like the docker image doesn't seem to log python output either!

Issue #55, opened 2024-05-08
Remove node-calls-python


josephjclark:
Sadly, `node-calls-python` isn't going to be sustainable fit for this project. It's a real shame.

The two issue are:
1) Logging - how to we get python stdout back into node?
2) Concurrency - how do we run multiple modules at once?

Admittedly questions are open about concurrency,  but at the very least it gives us problems with logging.

The problem is this:
* `node-calls-python` creates a single thread (so far as I understand it) in which all python executes
* I think that means that if one module starts blocking, all modules will die
* There is absolutely no threading/module control for npc. I get one context and I can't create more. Even if I could, I'm not sure
* I can't at the moment get the stdout for node-calls-python at all
* But if I had multiple jobs running, I could not tell the logs apart - I wouldn't know which was which

Possible solutions:
* I could use workerpool and create a new thread with `node-calls-python` for each job. But what's the point? We may as well just do `child_process.exec(0`

Basically I think that, for now at least, it's going to be just as easy to do `child_process.exec` and create a new python context each time. At this point I don;'t even care if that's a little slow in production - but I don't even have a reading on HOW slow.

TODO:

* Try and investigate blocking and queuing. Can I run two modules concurrently?  Most of the time it's just gonna wait for HTTP. I don't know enough about python to know if that wait is blocking. that should be a quick test. Well that wasn't easy but yes I think they run sequentially.

josephjclark:
Another issue coming up is that the python modules have to be DESIGNED to be long lived. You have to think very carefully about what gets instantiated where. Right now the logger has to be created inside main and given a unique id. Module-scoped loggers will forever log to the wrong file. And the same with any clients, you need to think about when it gets instantiated, and be sure you're not accidentally using the wrong client.

This is fairly hard engineering stuff. I'd actually PREFER that each python gets loaded, run and destroyed on every call - it's so much safer.

I suppose we could keep node-calls-python in always-reload mode, but in local dev that feels a little bit unstable (the server keeps dying and I haven't yet diagnosed why), but I also hope that some of the other libraries will make getting the stdout easier.

Issue #51, opened 2024-05-02
Enable websocket connections to stream logs through


josephjclark:
In order to send logs back to the client (and I'm thinking of the CLI in particular), the server needs to support a web socket.

Web socket requests need to be like:

* Client send an `upload-payload` event
* Server runs the service
* Server catches any stdio from python 
* Logs are send to the client via a `log` event
* The final result is sent to the client via a `complete` event

As I understand it, we can enable websockets on the same endpoints as the regular services - it's up to the client to add the upgrade payload header if it wants to upgrade to a websocket.

I'm a little worried in the differences in the interface but I think it should be managable - it's just two differen wrappers around the bridge really (plus an API to trap logs)

josephjclark:
Maybe `node-calls-python` is hurting me here. If  just shell out with `process.exec` I can at least trap the stdout

josephjclark:
If we're using node-calls-python and running two python scripts in the same thread concurrently, how do we distinguish the stdouts and send them to the correct process?

We don't want to divert ALL the stdout back to the caller. We just want to divert the stdout for that one module.

Maybe there's a wider issue here and I need to think a bit more about concurrency.

Again, this would all be simpler and more stable if I did a `child.process.exec` on the python scripts.

josephjclark:
Alright, this has bubbled out into: https://github.com/OpenFn/gen/issues/55

PR #56, opened 2024-05-09
Support websockets and redirect python stdout to the socket

josephjclark:
I've opened this as a separate PR to the main  server development stuff.

The main thread works fine, but when you call out to it from the CLI  you don't see any of the internal python logging (see #51). Not the end of the world but it does make for a pretty poor UX when calling the server from our lovely CLI.

The plan then is to use a websocket instead of a regular post to the service, and have the websocket stream results down. So the server now supports a HTTP or WS interface and you can connect to any service with either one. This all works great.

## The Difficulty

The problem is that python's stdout, although it'll be displayed in the server's log stream, doesn't appear in bun's stdout. So when python says `print()`, it goes to a different stream than buns `stdout` stream. Which means bun cannot trap it to send into the websocket.

This is a bit if a bummer. Without redirection support from `node-calls-python`, it means we have to use a child process to  get the stdout. And even if we did that  we might have to run some filters on the stdout, or find a way to make sure we're capturing the logs for the right job.

## Solution

My solution, which is pretty nasty, is to redirect python logs to file and read that file back into node.

It goes like this:

* I provide a utility API to get a logger
* the logger will log both to a file and to stdout
* Each "run" of a service will be given a unique path to log to
* bun will listen for writes to the log file, pull out the latest, and launch it out of the websocket
* When the job is done, it writes a special log line to tell bun that no more logs are coming
* Bun will then resolve the promise and return the result (AFTER all the logs have run)
* The log file is then removed

The worst bit of the solution is that the only way I've found to read the log back into node is this:

```
const child = spawn("tail", ["-f", logfile]);
  const rl = readline.createInterface({
    input: child.stdout
  });
  rl.on("line", (line) => {
    if (line === "***END***") {
    // ....
  });
```
Yep, I'm spawning a child process anyway.

## Failed Attempts

The solution is probably the dumbest and worst solution I could fine. But it kinda works.

Here are other things I tried:

* `node-calls-python` has no means of sending an IPC message between the python thread and the main thread. I suspect there's a way to do this but I don't think the maintainer has any interest and I do not have the knowledge.
* I couldn't really work out any other way to send a message from the python process to the main node process
* I did not try to emit the logs via HTTP. If the file logger is dumb and overcomplicated, then this seems dumber and overcomplicateder.
* Bun and node both have file streaming APIs, but neither is designed for live/continuous reading. They'll stop streaming when they hit the end of the file.
* Bun has a beautiful shell API, and instead of using node's child process (which remember is the very thing I'm trying to avoid with `node-calls-python`), I could do ``$`tail -f file.log` ``. But that also has no streaming interface - you have to await it, and of course `tail -f` won't return the promise.

Bit frustrating really.

## Issues

Well, if this works we can close:

#55 #51

## Further Work

We could probably release this when done... but I think there is a high chance that logs will interfere if two services happen to be running concurrently. 

But I'd rather add either (a) a very simple blocking queue or (b) plug in workerpool and ensure each python context only runs one thing at a time.

I don't think b) is a huge amount of work and probably would need doing anyway at some stage.


josephjclark:
We need to ensure that the logs from python are visible in the docker stdout. I think. And if not work out where/how to redirect them

josephjclark:
This has got me thinking AGAIN about the long running python idea.

Right now, every python module creates a logger when it is inited:

```
logger = createLogger("code_generator.prompts")
```

This is then shared through the module scope through the module's lifetime.

This means that multiple calls to the same service will use the same logger instance. Unless the logger is re-initialised in main and passed through to each function.  Would it get properly GC'd after the run or will references remain? Anyway - this feels like an awkward pattern and just the kind of constraint I don't want to put on devs.

As a result of this PR, that's a real problem, because each logger needs to be re-intialised to write to a different file. Ok, maybe main can do `logger.initialize()` and we can update the output handler, but this does feel kinda messy.

It's just the sort of problem that doesn't exist if we just spawned and destroyed the python process each time.

I could code a solution around this. But I'm realising that this approach makes contributors have to think about their python in a certain sort of way. What code needs to be re-initialized each time main is called? What secrets could potentially be held in memory, or maybe run stale? I just don't want our contributors worrying about these heavy engineering things.

So that might be it. I am 95% sure we're removing node-calls-python and just running a child process every time.

The websocket streaming stuff needs to stay, though.

PR #74, opened 2024-06-17
code_generator: fixing logger error

### LINKED ISSUE #70: code_generator: Logger calls for an unknown attribute 'generate'.
The current code generation service uses `logger.generate` to create logs. On `openfn apollo code_generator tmp/data.json -o tmp/output.json` it returns an error message:

Screenshot 2024-06-10 181421

A simple fix would be to use `logger.info` or `logger.debug` to avoid this error and print logs successfully.

SatyamMattoo:
Fixes #70 

Fixes an error caused by a typo by using logger.info instead of logger.generate in code generation service.


josephjclark:
Thank you!

PR #94, opened 2024-08-19
Claude

josephjclark:
naive integration of Claude into the job_chat service

This should work seamlessly with Lightning, although the Lightning env will need updating to use an Anthropic API key (we may want to rename the var in the code base too)

At the moment the token usage is going to be pretty high - we're just throwing everything into the prompt and crossing our fingers.

There are other things I want to look into here but I may spin them out into other issues:

* [x] Put more stuff in the system prompt (are there benefits apart from caching?)
* [x] Use prompt caching for the system prompt (this is a whole thing) https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching
  - adaptor APIs and the job writing guide would be good prompt candidates
* [x] Training the style of Claude's response. Verbose, terse, code only?
* [x] Careful use of [examples](https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/multishot-prompting#example-analyzing-customer-feedback)
* [ ] Optimise long prompts with <document> tags (maybe with rag?) https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/long-context-tips

Some other future work to consider:

* Use RAG to produce smaller inputs from the adaptor API. Re-using the search branch could give us good results here, resulting in smaller inputs and lower cost
* Use tools (functions which can call out to external services) https://docs.anthropic.com/en/docs/build-with-claude/tool-use. This basically means claude's response will be "hey call this tool for me", and you can send the result back to claude, which will then generate a final response. You could use it  to call for a live weather forecast and then generate a prompt off the back of it.  How  could we use this?


josephjclark:
This is disappointing:

Issue #109, opened 2024-11-13
Epic: Vocabulary Mapping Service


josephjclark:
# Overview

A service which generates a list of mapping for some source strings to some mapped vocabulary values - like a SNOMED or LOINC code.

This issue is a design for a general mapping service - but we'll be starting with a more specific use-case.

# Issues

* Create third party rag solution on apollo #111
* Embed SNOMED #114
* Embed LOINC #115
* Build a clean list of commcare input strings #113
* Set up a basic mapping service #112
* Optimise the mapping model #116
* Handle ambiguity (TODO - what happens if the model finds multiple mappings?)
* Validate results (TODO - how can we validate that the model's output is reasonable?)
* Feedback and guide the generation (TODO - how can we give the model feedback, corrections, and custom prompts?)
* https://github.com/OpenFn/apollo/issues/110

# Inputs

* Some list of string values to map. These must be in some kind of key: value format, where the key will be mapped to a value in the output. 
  * We don't have a formal design for the input yet. Likely we need unskilled users to be able to give us some kind of dump of input data. We should be able to support different input formats.
* Some list of target vocabularies or ontologies (ie, snomed). 
  * Because these vocabularies can be huge, can we take and map some kind of path?
  * Can we support custom standards?
* A template for output mappings - ie, a flat string, or a FHIR coding, or some other data structure.

# Outputs

A JSON object which can be used as a lookup table in job code.

The mapping will work something like this:
```
// This is our generated structure
const mappings = {
  'zero': {
    "system": "http://snomed.info/sct",
    "code": "105542008",
      "display": "Non - drinker"
  }
}

// Cross reference the key against our input data to get the mapped value
const mappedValue = mappings[input.answer];
```

# Design

All supported ontologies will be be embedded in a vector database

For each value in the input list:
 * Use similarity analysis to find possible matches in the embeddings (ie, load N possible matches from the database)
 * Call a model to figure out how the best mapping based on the shortlist of possibilities
 * Fit the mapped value into the correct data structure

Repeat this process until all inputs are mapped.

Return a JSON structure.

# API

The input payload should be something like:

* Inputs: key-value input values to map
* Hint: A custom prompt written by the user to guide generation
* Targets: list of target ontologies as mapping tagets
* Collection: { name, key } send the created data structure straight to a Lightning Collection

# Future Work

Handling uncertainty: If mappings are not clear, how can the service refer to a human for correction and validation?

[...]

PR #120, opened 2024-11-20
Add embedding service for issue 111

### LINKED ISSUE #111: Embeddings Service
We need to integrate a generic embeddings layer into Apollo.

At the moment, provide a search service which embeds the docsite - see #98 for outstanding work there.

But I'd like an architecture where any apollo service can access embeddings via a simple API

# API

The server needs an internal service like `embed`.

This should expose python services which can be called from the command line. It probably need an API like:

```
embed(datasetName, listOfChunks)
search(datasetName, query)
```

Where each chunk is probably a string.

We should not provide a direct REST API.

I think embeddings should be populated during the image build. So we need to work out a way for scripts to run at build time. Maybe every module can provide a build.py or build.sh which is automatically called by the build process

# Integrations

* The existing doc search is built on Milvus and Zilliz. It has a script to scrape the docsite and submit embeddings. It's a great reference and starting point, and when this work is done,  the search service should be ported over to it.
* The new mapping service will use this to embed vocabularies. It will probably be our first client

# Implementation Plan

The first step is to pick a suitable third party embedding service and database. There are many candidates for this.

The service should totally abstract the implementation, so that we can switch embedding services without affecting any of the dependent services

We should move all the vector DB and embedding client logic out of the docs embedding script. The docs script should call into the new embedding API, but not make any decisions itself about what database or embedding service to use.

Later, we can investigate the cost of hosting our own embeddings service. It would be super useful to handle user queries and support the AI assistant and stuff without calling out to a third party server.

hanna-paasivirta:
## Short Description

Create an embedding service to access/create a embedding database and search it with an input text.

Fixes #111 

## Implementation Details
This module is primarily built for the Vocabulary Mapping project #109, but it can be used in any future projects that require embeddings. It leverages the LangChain library to allow for easy access to different types of embeddings and embedding storage services.

The main functions are:
- `create_vectorstore()` -> Create a new embedding database collection with texts 
- `get_existing_vectorstore()` -> Access an existing embedding database collection
- `get_similar_texts()` -> Use an input text to query the embedding collection for similar texts

Currently the module only allows Zilliz as a vector store and OpenAI Embeddings as a model (both require credentials). New options can be easily enabled.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI


josephjclark:
Thank you @hanna-paasivirta ! This is exciting. I'll take a close look and see if I have any suggestions, but I'm keen for this to be a rough first cut which we'll iterate on over the coming weeks.

We'll need to break up #111 into some smaller issues to represent next steps (eg: port the docs importer into this embedding service, investigate and setup PG vector, etc)

services/embeddings/generate_embeddings.py
```diff
@@ -0,0 +1,182 @@
...
+    
+    return VectorStoreClass(embedding_function=_get_embedding_class(embedding), auto_id=auto_id, **init_kwargs)
+
+
+def get_similar_texts(input_text, vectorstore, search_type="similarity", search_kwargs={"k": 2}):
+    """
+    Retrieve similar texts from a vectorstore based on input text.
+    
+    Args:
+        input_text: Text string to use for similarity search
```
josephjclark, line 141: Ah I'm interested in this - so we don't need to manually generate embeddings for our search strings? Langchain does this for us based on the built-in embedding function?

This is an area I've been curious about.

josephjclark (commented):
This is great Hanna thanks.

I've left a couple of notes on comments and general organisation.

Let's have a chat in the morning about the main interface and how this service will actually be used. Subject to that conversation I might have more suggestions or we'll merge in and move forward :muscle:

services/embeddings/generate_embeddings.py
hanna-paasivirta: No we don't, although we could if we wanted to. It'll just use the embedding type associated with the vector store.

services/embeddings/demo_data/data.json
```diff
@@ -0,0 +1,59 @@
...
+"sender_name": "Mr. Darcy",
+"timestamp_ms": 1057190400000
+},
+{
+"content": "Come now, Mr. Darcy, a man of your consequence must employ nothing but the finest automation tools. Or do you still prefer manual data entry?",
+"sender_name": "Lizzy",
+"timestamp_ms": 1057190300000
+},
+{
+"content": "I must confess, the OpenFn integrations have transformed our estate management beyond measure.",
```
josephjclark, line 24: I love this.

services/embeddings/demo_data/data.json
```diff
@@ -0,0 +1,59 @@
...
+{
+"creation_timestamp": 1057188000,
+"uri": "pemberley_workflow_diagram_1803.jpg"
+}
+],
+"sender_name": "Mr. Darcy",
+"timestamp_ms": 1057188000000
+},
+{
+"content": "Indeed, Mr. Darcy. May your workflow automations bring you much efficiency.",
```
josephjclark, line 44: But what happens next??

services/embeddings/generate_embeddings.py
```diff
@@ -0,0 +1,138 @@
...
+        try:
+            EmbeddingClass = self.embedding_classes[self.embedding_type]
+            return EmbeddingClass()
+        except KeyError:
+            raise ValueError(f"Unsupported embedding type: {self.embedding_type}")
+    
+    def _get_vectorstore_class(self):
+        """Get vectorstore class based on the specified type."""
+        try:
+            return self.vectorstore_classes[self.vectorstore_type]
```
josephjclark, line 49: Out of curiosity, are these vectorstore classes expensive to instantiate?

Do they try to maintain a live collection to the DB or are they "lazy" and just call out when needed?

josephjclark (changes requested):
Hey @hanna-paasivirta 

I've left a few comments inline but the headline here is that the structure of this isn't what I was expecting after where we ended up in our chat yesterday.

The value of this branch is in setting up a _pattern_ that makes it easy for apollo users to create and search embedded datastores. That pattern will evolve as we move forward, but I want to get us off to a clean start. And more importantly, make sure we're on the same page and moving in the same direction.

To better explain my expectations, I've raised a draft PR #121 - please have a careful read through that and let me know what you think.

One other thing for when the code is done:  can I ask you please to create a simple, very high-level readme file which broadly explains what the embedding service is (or aiming to be), and how datastores are expeted to be used (we talked yesterday about the `stores/*`  folder structure and some kind of import of a search function). I'm looking for something really short, without much detail because things are likely to change quickly. Just an overview in maybe 100 words to give context to, for example, Elias if he were to look at this folder tomorrow and wonder what was going on.

Thank you :pray:

hanna-paasivirta:
In the latest commits I have:

- Changed sys arg in entry.py – I think there was an index typo that meant the output json path would not be read
- Changed the main function in embeddings_demo.py to reflect what entry.py expects (json input and output)
- Renamed from generate_embedings to embeddings for clarity and to match entry.py requirements
- Renamed create_collection to add_docs
- Removed the main function in embeddings.py as it will not normally be used directly from this folder
- Added READMEs
- not included the demo.py from PR #121 , as I think it does the same thing as embeddings_demo.py now.

hanna-paasivirta:
Got it, does it look right now? I moved the vector store set up and instance creation code inside embeddings/demo/example.py

services/embeddings/embeddings.py
```diff
@@ -0,0 +1,109 @@
...
+    def _get_vectorstore_class(self):
+        """Get vectorstore class based on the specified type."""
+        try:
+            return self.vectorstore_classes[self.vectorstore_type]
+        except KeyError:
+            raise ValueError(f"Unsupported vectorstore type: {self.vectorstore_type}")
+    
+    def add_docs(self, docs, drop_old=True, **kwargs):
+        """
+        Create a new vectorstore from documents and initialise it with the specified settings.
```
josephjclark, line 55: Is this comment still accurate? Looks like it's fallen out of date to me - I don't think this creates a new vector store at all :thinking:

services/embeddings/generate_embeddings.py
josephjclark: @hanna-paasivirta This question is still outstanding :) It affects how and when we initialise our vector stores.

services/embeddings_demo/embeddings_demo.py
```diff
@@ -0,0 +1,37 @@
+import os
+from dotenv import load_dotenv
+from embeddings.demo.example import get_demo_vectorstore
```
josephjclark, line 3: I'd really like a cleaner import here :thinking: 

How can we set up the package structure to support this?
```
from embeddings import demo

demo.add_docs()
demo.search()
```

That would be really neat. I was thinking about having a sub folder per file, which allows stores to have many supporting files and still give us a tidy folder structure. If that's in our way, maybe we have a top level entry point file (`example.py`), AND a subfolder to hide  away the messy stuff, if needed (`example/setup.py`, `example/data.json`)

We'd also have to do away with the `get_demo_vectorstore` and initialise the store when the module is loaded. Which, as I've asked elsewhere, I'm  not sure what the cost of that is.

If we really have to have an initialisation function, maybe we can just call it `connect`? Which I expect to give me a unique connection to the underlying database.

That would give me code like:

```
from embeddings import demo

store = demo.connect()
store.add_docs(...)
store.search(...)
```

josephjclark (commented):
Hey @hanna-paasivirta 

This is close, thank you!

I find it a little weird that the `demo` store has an `example.py`. I'd like to just use the word `example` or `demo` - and I think I prefer example because it's an example data set which is _used_ by the demo.  But this mixing of language is hard on my brain and I keep getting mixed up on what I'm looking at :exploding_head: 

The other big thing is that the data loading still isn't super clear to me, and I'd like to see a little bit of tidying up around there.

The idea goal of this PR is a really clear example and just enough documentation that I could ask someone like Hunter or Elias to add a new embedding store, or to port the existing docs embedding into this structure. I don't think we're quite there yet.

services/entry.py
```diff
@@ -37,7 +37,7 @@ def call(service, input_path, output_path):
     input_path = sys.argv[2]
     output_path = None
 
-    if len(sys.argv) == 5:
+    if len(sys.argv) == 4:
```
josephjclark, line 40: TODO for me: remember to test this with CLI

services/embeddings/embeddings.py
hanna-paasivirta: The LangChain function from_documents will create a new collection while adding documents. I'll edit the comment from "vectorstore" to "collection" to clarify.

services/embeddings_demo/embeddings_demo.py
hanna-paasivirta: It seems a little unexpected to me to do something potentially costly/unwanted (wiping a database) in the background on just an import. If it's ok, I'll just rename it to connect() to make it clear that something will happen when you use the class instance.

services/embeddings/generate_embeddings.py
hanna-paasivirta: This function just gets the LangChain class for a vectorstore. But in general, I'm not sure how to answer. I think the cost of instantiating a vectorstore in the code is negligible. The costly operations are generating embeddings, indexing, storing, similarity search, and insertion.

hanna-paasivirta:
I have just pushed a commit where I did the following:

- Moved example.py out of the demo folder onto the same level as embeddings.py
- Renamed get_demo_vectorstore to connect
- Changed to import whole module instead of specific functions
- Edited the comments you noted
- Added the demo and an explanation of its structure in the README of embeddings

Did I miss or misunderstand anything?

services/embeddings_demo/embeddings_demo.py
josephjclark: Yeah agreed!

`connect` should just give me a live connection to a database, which I can then do as I want. And it's fairly clear from that semantics that that's what I expect. So that's cool!

One note on this: in the example store, the `connect` function will also load data. Which is a little funky because if two  different services connect, they'll duplicate the data load.

Maybe just for the purposes of the demo, we can have a `connect()` and a `load_default_data()` function in the example store, and then the demo script will explicitly connect AND load the data?

This is also a  bit weird - we'd usually have some offline process to populate the store and not make it part of apollo like this. But it keeps the data loading logic in the store (which I like) and takes side-effects out of `connect` (which I like). So I think it's an improvement?

We'll have to look more closely at when and how the embedding stories get loaded. But we'll defer that here - we're going to bump into it pretty quickly on your mapper!

hanna-paasivirta: I added a load_default_data() in the demo – does it look like what you meant?

josephjclark: Perfect!

josephjclark (approved):
Thank you @hanna-paasivirta ! :tada: 

Just one last thing from you please (sorry I should have mentioned this earlier)

Can you run `bun changeset`  and follow the prompt to generate  a changeset file with a vague user-facing summary of what this PR does. Something as simple as  `Add an embedding service` will suffice for now.

This should be a minor change, because we're adding something new.

Then commit and push the file to here,and I can merge.

Happy to do this live on the call with you tomorrow morning :)

services/embeddings/example.py
```diff
@@ -0,0 +1,32 @@
...
+    """Initialise the vector store instance to be used by the embeddings_demo service."""
+
+    load_dotenv()
+
+    store = VectorStore(
+        collection_name="demo_project",
+        vectorstore_type="zilliz",
+        embedding_type="openai",
+        connection_args = {
+            "uri": os.getenv('ZILLIZ_CLOUD_URI'),
```
josephjclark, line 16: Oh @hanna-paasivirta please make sure this uses the same env var as the basic search service :pray:

PR #123, opened 2024-11-27
add cors support

josephjclark:
Support cors in requests. Needed for the docsite helper


josephjclark:
Are there security implications of this change that I'm not understanding?

PR #124, opened 2024-11-27
Docsite explainer

josephjclark:
## Short Description

Details to follow

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/services/docsite_explainer/README.md
+++ b/services/docsite_explainer/README.md
@@ -1,0 +1,31 @@
+# Docsite Explainer
+
+This is a service for explaining passages in the OpenFn documentation.
+
+This service uses an input text to query a Claude model for a clarification. It also searches an embedding collection containing the full documentation to give the model passages from it as additional context.
+
+## Implementation Notes
+
+This service currently uses the Zilliz embedding storage provider for OpenAI embeddings, and requires both Zilliz (URI, Token) and OpenAI credentials to run.
+
+## Usage
+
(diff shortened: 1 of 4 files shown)


josephjclark:
On top of the other work needed here, we need to pass a responsible AI wand over this (here or on the docs side?)

josephjclark:
@hanna-paasivirta Can I ask you, when you have time, to:

1) Re-base this branch against main (its a bit behind) (I can help you with rebasing)
2) Write a _brief_ PR overview and list what needs to be done before we can release to production (excluding the build-all-the-things thing)
3) Propose a solution to build an in-memory cached explanation of every line (I think we should develop this on a separate branch, just in case we decide not to deploy it)

hanna-paasivirta:
Proposed solution to build an in-memory cached explanation of every line:

One option to consider as part of the cost evaluation and front-end design is to build an in-memory cached explanation of every possible input paragraph for each version of the Documentation. On the backend, this would need the following process to run once:

1. Script to split the documentation into the correct paragraphs. Save and index.
2. Feed the splits to the document explainer pipeline. (Could use the Anthropic Batches API to process at non-peak times to lower costs).
3. Save the answers on the server. Index so they can be fetched with the corresponding input.

In this scenario, the front-end "explain" button would not trigger the document explainer pipeline, but would fetch the relevant saved answer from the server.

josephjclark:
Fantastic write up @hanna-paasivirta , thank you!

PR #126, opened 2024-11-30
One Error Message To Rule Them All 🧙🏽‍♂️ 💍 

### LINKED ISSUE #125: Standardise error messages
We should create a standard error class in python land for all apollo errors. Python services should return this if there's an error, which enable the webserver to return a proper error code and standardised payload back to the client.

This should serialize into a standard format that the web server can use.

I'm thinking about a low like this:

* A request comes into the bun webserver
* The webserver calls into pythonland via the bridge, passing a JSON payload as a dict
* The python module does some stuff - but there's an error! Maybe Claude hit a rate limit and our request failed
* Python creates an error object, like `ApolloError(500, type='RATE_LIMIT', message = 'Rate limit exceeded, slow down pal'),`. Or something
* The python module returns this object from its main, serializing to JSON
* The webserver takes the result and says: is it shaped like an error object? Of a form like `{ errorCode: 500, errorType: '', errorMessage: "'' }`
* If so, I'll return a a HTTP error status code (from `result.errorCode`) with some error details in the payload as JSON
* Otherwise I'll return a 200 status code with the payload in the json.

I don't think we need a standard wrapper for successful results. The idea is that the python  returns whatever it wants. We'll just add this new convention for errors, which makes it easier for the web server to set the status code in the response.

elias-ba:
## Short Description

This PR introduces a standardized error handling mechanism for Python services, enabling consistent error responses and proper HTTP status codes.

Fixes #125 

## Implementation Details

The implementation provides a simple way to handle errors in Python services. Any response with a numeric `code` field will be treated as an error and returned with the appropriate HTTP status code.

Python services can now raise errors using the `ApolloError` class:

```python
from util import ApolloError

def main(data: dict):
    if rate_limit_exceeded:
        raise ApolloError(
            429,
            "Rate limit exceeded, please try again later",
            type="RATE_LIMIT",
            details={"retry_after": 60}
        )
    return {"success": True}
```

The error handling is centralized in `entry.py`, which catches both `ApolloError` and unexpected exceptions, converting them to a consistent format. Services don't need to call `.to_dict()` - they can just raise the error.

Example usage in `job_chat.py`:
```python
def main(dataDict):
    try:
        data = Payload(dataDict)
        result = generate(data.content, data.context)
        return result
    except anthropic.RateLimitError:
        raise ApolloError(
            429,
            "Rate limit exceeded, please try again later",
            type="RATE_LIMIT",
            details={"retry_after": 60}
        )
    except Exception as e:
        raise ApolloError(500, str(e))
```

Tests in `platform/test/server.test.ts` verify the error handling for rate limits, unexpected errors, and successful responses using a dedicated test service.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


elias-ba:
Hey @josephjclark, I had really good progress on this work so far, I don't think it is ready yet tho, but here are my current blockers and things I would like to talk with you.

1. Test Setup Issues
The tests I wrote for the new error handling service are failing, likely due to dependency/setup issues. When running `bun test`, I get errors that seem environment-related rather than logic-related. I can either investigate this further tomorrow or we could pair on it to resolve these issues faster.

2. Implementation Strategy
I've included an example in the PR description showing how we could implement this error handling in the `job_chat` service. This should give a clear picture of how we can gradually roll this out to other services. Would love your thoughts on this approach before we proceed with wider implementation.

Let me know if you'd like to pair on resolving the test issues or if you have any feedback on the implementation strategy.

josephjclark:
@elias-ba let's chat on Monday!

platform/test/server.test.ts
```diff
@@ -83,4 +83,46 @@ describe("Python Services", () => {
       });
     });
   });
+
+  describe("Error handling", () => {
```
josephjclark, line 87: TODO for me: 
1. Test this locally
2. Think about whether `test_errors` is the best approach

josephjclark:
@elias-ba I'll think about the test and get back to you :)

elias-ba:
@josephjclark, please have a critical eye on the PR, as I changed the createLogger func name to create_logger and thus changed it in all files where it is referenced. I have verified and everything seems okay to me, but as my tests are not working yet, it's a bit scary

platform/src/util/errors.ts
```diff
@@ -0,0 +1,15 @@
+export interface ApolloError {
+  errorCode: number;
+  errorType: string;
+  errorMessage: string;
+  errorDetails?: Record<string, any>;
+}
+
+export function isApolloError(value: any): value is ApolloError {
```
josephjclark, line 8: Can we just say `if (value.statusCode)` ? Then any object coming out of python with an errorCode will be treated as an error

I think that's simple and robust enough for our needs

josephjclark: You can keep this util function though

josephjclark: Just simplify it

elias-ba:
@josephjclark all is good now. You can merge and I will branch off to start working on the Job Chat service. Thanks

josephjclark:
Tests are passing! Hurrah!

PR #131, opened 2024-12-06
Release: embeddings

### LINKED ISSUE #111: Embeddings Service

josephjclark:
## Short Description

Release branch for  the embeddings service

Closes #111


josephjclark:
I think this is ready to go? Embeddings isn't used by anything so its nice and low rish.

I'd just like to run a few tests on monday to make sure job chat works with lightning from this branch. Then we can release and deploy :ok_hand:

josephjclark:
Dang, the arguments change in entry.py broke production. I've reverted and fixed it on main. I'll raise an issue to look at this.

Issue #115, opened 2024-11-13
Embed LOINC


josephjclark:
We need a solution to:
* Download a list of LOINC vocabulary values
* Maybe convert them into a standard internal JSON representation (so that LOINC and SNOMED vocabs look the same)
* Embed and load them into our vector database

Questions include:
- How much of LOINC do we need? Can we identify useful subsets and just load those in?

josephjclark:
See huggingface: https://huggingface.co/datasets/awacke1/LOINC-Clinical-Terminology

Issue #114, opened 2024-11-13
Embed SNOMED


josephjclark:
I think LOINC is a higher priority, but still

We need a solution to:
* Download a list of snomed vocabulary values
* Maybe convert them into a standard internal JSON representation (so that LOINC and SNOMED vocabs look the same)
* Embed and load them into our vector database

Questions include:
- SNOMED is proprietary and you need to pay for access. Is there any way we can get a list? A partial list is fine for test purposes
- SNOMED is huge right? Can we break it up into ontologies and only embed some of them?

Because snomed is paywalled we think its value is limited to us.

PR #134, opened 2024-12-11
Add pinecone store functionality to vector stores

hanna-paasivirta:
## Short Description

Add Pinecone vector store access and threshold-based similarity search to our VectorStore class. Then, embed LOINC and SNOMED datasets (#114 and #115)

## Implementation Detail

Add Pinecone access to the VectorStore class, which currently only allows Zilliz stores. Allow use of threshold-based similarity search through Pinecone, in addition to the existing top k search.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI


hanna-paasivirta:
Added LOINC and SNOMED embeddings to Pinecone. This PR still needs to add a README.

The embedded datasets may need further work as we develop the project:
- One tricky aspect is knowing which would be the best columns to embed for similarity searches, which ones to leave for rule-based filtering, and which ones to feed in as extra context to the LLM after retrieval. The decision requires some familiarity with the dataset to avoid unexpected results because there’s different types of data and empty cells. While some column values might potentially be misleading and give too many similarity hits (e.g. maybe LOINC “SURVEY.MTLHLTH” values in the Class column might give false hits for other “Survey” types, but the class “ALLERGY” values could be essential context for the similarity search for entries like “Goat cheese IgG”.
- The embedded part and additional columns might need to be edited and the texts re-embedded in future versions. We might also consider using an LLM to add our own custom explanatory columns, which can be included in the embedding or in the additional context given to the LLM.

services/embeddings/health_mappings.py
```diff
@@ -0,0 +1,110 @@
+import os
```
josephjclark, line 1: This file needs moving into `services/embeddings/health-mappings`

Also, I was actually expecting one folder for `loinc` and another for `snowmed`. One folder per store, rather than lumping them together.

The idea is that loinc and snomed are stores that can be freely searched by other services, and not innately bound to the healthcare mappings project.

services/health-mappings/embed_loinc_dataset.py
```diff
@@ -0,0 +1,20 @@
...
+
+load_dotenv(override=True)
+
+def embed_loinc_dataset(input_query):
+
+    # Initialise the vector store instance
+    store = health_mappings.connect_loinc()
+
+    # Embed and uplaod data to the vector store instance
+    health_mappings.upload_loinc_data(store)
```
josephjclark, line 13: I think I'd expect the logic for the upload (`upload_loinc_data`) to be in this script, so that all the upload logic is self-contained here.

If you're just using loinc as a client/consumer, you don't care - and don't need to care  - how the data is uploaded. All you need to care about is what search functionality is exposed.

josephjclark (commented):
Thanks for this @hanna-paasivirta ! I know it's a work in progress but I've added some comments for things that lump out at me.

We'll have to talk more about Pinecone next week.

One thing that strikes me - and I know you're not finished - but I can see:
- code to connect to pinecone
- code to embed data in pinecone

But no code to use the store to search pinecone.

Do you have any test code to support this work? I don't mean formal unit tests - I just mean any scripts or notebooks you're using to validate your work as you go.

All your messy working files should be checked in to your PRs. Not only does this give you a backup stored online, but this stuff validates your work and enable others to use it more easily.

services/embeddings/health_mappings.py
```diff
@@ -0,0 +1,110 @@
...
+    store = VectorStore(
+        collection_name="snomed_mappings",
+        index_name="apollo-mappings",
+        vectorstore_type="pinecone",
+        embedding_type="openai"
+    )
+
+    return store
+
+def preprocess_loinc(df, keep_cols, embed_cols):
```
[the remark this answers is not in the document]
josephjclark: After talking it sounds like this is very doable - but lets wait until we import ICD 10 and handle it then :)

hanna-paasivirta:
Changes added so far:
- Upload notebook
- Split LOINC and SNOMED into separate folders for easier access for different projects
- Add search example to the upload modules

I still need to test the modified code (ideally without uploading all the data again), and come back to the vectorstore arguments organisation options.

hanna-paasivirta:
Changes:

- Renamed project names and the main functions to match entry.py expectations 
- Did not nest folders as discussed because it's difficult to achieve with the current project path setup in entry.py. Changing entr.py in turn might lead to us having to change other projects. This project is probably not mature enough for us to decide on changing all projects based on it, and the current folder structure works well enough I think.
- Further testing done, all should work well (tests can be viewed in the notebook in `embed_loinc_dataset`)
- Added READMEs

The next steps will involve using the embedded datasets for RAG. The datasets will be accessed through a separate module which will be built in another PR.

josephjclark (approved):
This looks great, thanks Hanna

Just a note that I'm still not sure the right way to invoke the scripts that load embeddings data. I've created a separate issue for that which we'll look at in the new year https://github.com/OpenFn/apollo/issues/136

PR #135, opened 2024-12-12
Refactor services entry point to use named arguments

### LINKED ISSUE #132: Fix arguments in entry.py
`entry.py` is the entrypoint to python modules from Javascript land. It's used by the REST API at runtime, as well as by commands like `bun py`  in local dev. One of its important jobs is to ensure that relative module paths work.

There's a problem with arguments in the main function though. We use 4 or 5 positional arguments - depending on the calling context - to work out what the different flags are.

I think at the moment, we call 5 arguments from the web server. But if you call entry.py directly from python it only takes 4 arguments.

Whatever  - the interface is absolutely awful, and this is a key part of the server infrastructure.

So I'd like to re-write the arguments to entry.py to use named arguments (ie `--output-path`) rather than relying on positions. That should make life easier for everyone.

The server's unit tests WILL  fail if the arguments are wrong. We should probably ensure a python facing test too

elias-ba:
## Short Description

Refactor Python service entry point to use named arguments instead of positional arguments, making it easier to use, more maintainable, and less error-prone.

Fixes #132

## Implementation Details

The Python service entry point (`entry.py`) is a critical part of our system, used both at runtime by the API and during local development. Previously, it relied on positional arguments for service configuration, which caused confusion and made the script harder to maintain.

### Key Changes
1. **Switch to Named Arguments**  
   The script now uses named arguments via `argparse`, making it more explicit and self-explanatory:  
   
   ```bash
   python services/entry.py <service> --input input.json --output output.json --port 5000
   ```
   
   - `<service>`: Required; specifies the service to run.
   - `--input` (`-i`): Optional; path to the input JSON file.
   - `--output` (`-o`): Optional; path to the output JSON file (auto-generated if not provided).
   - `--port` (`-p`): Optional; sets the Apollo server port.

2. **Improved Error Handling**  
   - Clear error messages are provided for common issues like invalid JSON input or missing files.
   - Type hints have been added for better clarity and maintainability.

3. **Integration**  
   - `bridge.ts` has been updated to use the new named argument format when spawning Python processes.

These changes enhance the usability and maintainability of the entry point while maintaining backward compatibility for existing API.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/platform/src/bridge.ts
+++ b/platform/src/bridge.ts
@@ -30,38 +30,16 @@ export const run = async (
     console.log("Initing output file at", outputPath);
     await Bun.write(outputPath, "");
 
-    // const proc = Bun.spawn(
-    //   [
-    //     "poetry",
-    //     "run",
-    //     "python",
-    //     "services/entry.py",
-    //     scriptName,
-    //     JSON.stringify(args),
-    //   ],
-    //   {
-    //     onExit: async (proc, exitCode, signalCode, error) => {
-    //       // exit handler
-    //       const result = Bun.file("out.json");
-    //       const text = await result.text();
-    //       resolve(JSON.parse(text));
-    //     },
-    //   }
-    // );
-
-    // Use nodejs spawn
-    // I seem to have to use this because the bun stream doesn't work with readline
-
     const proc = spawn(
       "poetry",
       [
         "run",
         "python",
         "services/entry.py",
-        scriptName,
-        inputPath,
-        outputPath,
-        `${port}`,
+        "--service", scriptName,
+        "--input", inputPath,
+        "--output", outputPath,
+        "--port", `${port}`,
       ],
       {}
     );
--- a/platform/src/bridge.ts
+++ b/platform/src/bridge.ts
@@ -36,14 +36,10 @@ export const run = async (
         "run",
         "python",
         "services/entry.py",
-        "--service",
         scriptName,
-        "--input",
-        inputPath,
-        "--output",
-        outputPath,
-        "--port",
-        `${port}`,
+        ...(inputPath ? ["--input", inputPath] : []),
+        ...(outputPath ? ["--output", outputPath] : []),
+        ...(port ? ["--port", `${port}`] : []),
       ],
       {}
     );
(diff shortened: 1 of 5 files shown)


josephjclark:
Thank you so much @elias-ba ! I look forward to reviewing this tomorrow

platform/src/bridge.ts
```diff
@@ -30,37 +30,19 @@ export const run = async (
...
-    // Use nodejs spawn
-    // I seem to have to use this because the bun stream doesn't work with readline
-
     const proc = spawn(
       "poetry",
       [
         "run",
         "python",
         "services/entry.py",
+        "--service",
```
josephjclark, line 39: Can we make script name required and position? I think the absolute minimum here is `python services/entry.py <service>`, and everything else is probably technically optional (even if we usually use it)

josephjclark (changes requested):
Thanks @elias-ba! This seems to work great. One request please, then I'll ask Hanna to test (this works great with the assistant though)

One question: how do we use `entry_test.py`?

elias-ba:
Thanks for the review, @josephjclark! 🙌 

I’ve implemented the changes you requested, so everything should be good to go now. Let me know if there’s anything else you’d like adjusted.

You can run the `services/entry_test.py` using `pytest` like this:

```bash
pytest services/entry_test.py
```

I have updated the `pyproject.toml`file at the root of the project to include `pytest`. It's a cool little package for writing and running unit tests in Python. So you may probably need to install it by running `poetry install` before:

Let me know if you hit any snags or need help with it! Appreciate your feedback, as always! ✌️

josephjclark:
Thank you @elias-ba :pray: 

@hanna-paasivirta when you're next online:

Remember that change you made to the arg parser in https://github.com/OpenFn/apollo/pull/131/commits/b383b0d4364855689a501f2a16037f639a68008d ?

Could you repeat your test to expose that? It'll fail on main now but should work against this branch. If you have trouble please let me know ASAP. Thanks!

hanna-paasivirta:
@josephjclark It seems to work fine now. I modified the example command in embeddings demo to match the new named arguments.

josephjclark:
Great catch @hanna-paasivirta , thank you!

Issue #112, opened 2024-11-13
Basic Mapping Pipeline


josephjclark:
The first mapping service, Aisha's mapping service, is designed to be the simplest form of mapper we can get.

The mapper is designed to generate a single sheet of the lookup tables spreadsheet. To get all the lookup tables needed for the asri-satusehat project, you have to run the mapper multiple times.

It'll work like this:
* input: keys - a list of commcare strings (keys), perhaps with context. As a file, one key per line
* input: vocabs - a target vocabulary (one item per line) to which we want to map. Maybe as a tuple-like system/display/code
* output: a JSON object with commcare keys mapped to fhir objects

And that's it.

The payload needs to be like:
```
{
  keys: ["Typhoid fever", "Dysentery, bacterial", ...],
  vocab: ["A01.0", "A06.0"]
}
```
(this example maps to ICD10 codes, not fhir)

We will want to add some script or CLI support to build this payload from files (the openfn CLI could help here)
# Design Notes

In this system, do we even need a database and embeddings??  We're asking the human to slice down the dataset for us so we may not even need rag. Having said that, a) If we run the model for each individual mapping, then maybe we still want to use rag even from a small list of inputs, and b) if we build it with rag now, then we can scale better later.

Ok, so we probably do want to use embeddings. But I think we still need to dynamically embed the inputs from the user, rather than pulling from a larger corpus.

This is a tricky detail and key to the whole design.

# Mapping Notes

Which sheet do we want to start with? They all have problems...

- diagnosis with body site - there are two things to map: ICD10 codes and snomed body sites in FHIR
- observations - map strings to a category AND a loic code
- option sets - commcare strings to snomed codes

# Validation

In Aisha's project specifically, we have the actual answers to the mapping in the spreadsheet.

We should be able to use that information to build formal validations of the results. Like we can literally diff the output and report a success score

Maybe, more generally, if you're doing some mapping and you get some results out and verify them, but then you want to go back and regenerate the mappings because you've added new inputs (or one turned out to be wrong). You might well want to pass your previous result as a validation set to make sure things aren't moving

# CLI Integration

I don't know if we want to integrate this into the CLI as part of this work, but I think it's helpful to look ahead and think about what the CLI interface might look like

```
openfn map-vocab \
	--vocab path/to/file \    # or pre-loaded dataset name
	--inputs path/to/file \
	--prompt "map the inputs to fhir codings like { system, display, code }"
```
# Limitations

Obviously this is a very limited system!

[...]

PR #137, opened 2024-12-20
Vocab mapper

hanna-paasivirta:
## Short Description

This PR adds our first mapping pipeline prototypes for generating some of the mappings of the asri-satusehat project. This involves mapping CommCare terminology to LOINC and SNOMED terminology according to the guidelines of this particular project.

See more details in issue #112 

## Implementation Details

This exploratory research for the vocab mapping pipeline is contained in the notebook `vocab_mapping_pipeline.ipynb` . The notebook `vocab_mapper_test.ipynb` provides a quick way to test the approaches with different inputs.

The first result of this notebook is to identify two distinct mapping tasks in our project:
a) Similarity-based mapping tasks – e.g. mapping CC terminology like "pulse_bpm" to  LOINC terminology "Heart rate"
b) Tagging-based mapping tasks  – e.g. diagnosis mapping to their corresponding body parts ("Measles" -> "Skin structure")

The second result of this notebook is the testing of three different pipelines for these two tasks:
1) Similarity search first method – We search the vector store (LOINC) for potential matches. We give the matches as additional context to an LLM and ask it to output the correct LOINC term.
2) Similarity search second method – LLM outputs what it thinks the term is, then we use it to retrieve results using hybrid search (similarity search + keyword search).
3) Tagging task – LLM outputs what it thinks the appropriate term would be (SNOMED body site), then we use it to retrieve entries from the vector database.

There are more details and some error analyses in the notebook.

The results for the tested mappings from the project can be found in [these sheets](https://docs.google.com/spreadsheets/d/1TFXQ1XsdfhUgNuVHteBknn_w_isTBm-6g0QTLTHsk5A/edit?usp=sharing).  The sheets additionally test a Claude Projects-based approach, where we add the LOINC/SNOMED datasets as Project data (circumventing the prompt token limit) and prompt the model to perform the mapping task.

As part of our preliminary research, the notebook tests zero-shot mapping with just brief input terms (e.g. "gdp") with no additional context. In some cases, the mapping is impossible to complete with no additional context, but these results will form our baseline results.

For similarity-based mapping using LOINC, Claude Projects had an accuracy of 12/53, while my best RAG pipeline had an accuracy of 11/53. The accuracy is low for all methods when looking at perfect results. However, many answers seem very reasonable and I need to understand why they are incorrect (are they incorrect only in this project, or objectively incorrect?) and what additional guidance we could expect to have access to and how to feed that to the model.

Some key bottlenecks we can see in the results: low accuracy of embedding search with dozens of nearly identical CT entries; lack of overall project instructions (e.g. when to pick the "narrative" or "structure" version of fields); lack of essential context (e.g. "how many children" maps to "births.still living", or "syphilis_result" maps to a test by "Hemagglutination" rather than other methods).

The experiments in this notebook will help us identify the best approaches to develop further. The promising results using Claude Projects and its agentic pipeline should also guide us towards developing a reliable mapping tool.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI


hanna-paasivirta:
As mentioned above, the model gives many reasonable mappings which differ from the target mappings. To see how we should proceed, I went over some [examples](https://docs.google.com/document/d/1m5vCNQpxAF_o9ZgyTUm5ejDBv5HwDKU_G_bTPPiJJXg/edit?tab=t.0) with Aisha from the implementation team. The key takeaways for our evaluation are that a) many of these mappings are necessarily educated guesses, and so there may be several acceptable answers, and b) there is sometimes extra context that determined the choice of a mapping.

Based on this, I re-evaluated & double checked the results. Focusing on the synonym issue, specifically, our 2nd approach in comparison to the Claude Projects approach:

- The pipeline I created has an estimated 23/53 (**43.4%**) correct, while Claude Projects has 22/53 (**41.5%**). (This is estimated by me in this development phase, not a medical expert.)
- Recap of differences between the two pipelines:
  - Claude uses an agentic approach - i.e. it makes its own decisions of how to proceed (runs code to parse the datasets etc). It has to rely on long-term memory. (+ hidden steps from Anthropic that we don’t know about that should increase accuracy).
  - Our pipeline uses an explicit LLM chain approach where we decide exactly what the steps are. The aim of this is to constrain the answers to reduce hallucinations and ultimately achieve higher accuracy. We run it on one input at a time, to reduce reliance on long-term memory.
- The accuracy scores between the two systems are still almost the same. However, there’s now **a clear difference in where the majority of errors are coming from.** In Claude Projects, there’s at least 15 errors that are **hallucinatory** (changed codes mostly, or names). In our pipeline, there’s no hallucination issue. Instead, there’s about 14 error cases where **the pipeline fails**, i.e. the model says it does not know the answer (different possible reasons – probably wrong original guess, or the search results are not broad enough). Given my sloppy first pass at this type of pipeline, I think the result is very promising. 
- Hallucination is harder to spot than an error where the pipeline fails. It may be preferable to pick a system that does not hallucinate, even if the accuracy were lower.
- Hallucination might also be harder to fix. We could output Claude Projects results into a sheet and then develop our own approach focused on identifying and fixing the hallucinations. But we would then be building a pipeline based on a UI product that might become inaccessible. It is safer to develop a pipeline based on plugging in any multipurpose LLMs.
- There is a systematic error with the "narrative" versions of LOINC entries - this seems fixable based on Aisha’s answer. If we fix this, **the estimated accuracy of the Claude Projects approach is 23/53 (43.4%**, increases only by one point, because most of these cases also have a hallucination problem), while **our pipeline’s accuracy would be 28/53 (52.8%)**. This is estimated based on counting outputs that give the non-narrative version of entries correctly without hallucinations.

To improve our very basic pipeline’s accuracy, we should aim to address:
- As mentioned above: cases where the model says it does not know the answer (wrong first pass guess, or the search issue).
- Remaining errors besides those mentioned above: 
  - Some are due to lack of additional info (from project or extended LOINC dataset with explanatory notes like on the website); 
  → See if website metadata can be added
→ Process Aisha’s note file and additional documentation
  - A few are too specific; 
→  Adjust prompt
  - A few are more obvious bad reasoning
→ Adjust prompt; 
→ Second judge LLM (leave for later)

I'm going to aim to address these, while keeping our new idea for what the final v1 mapping tool could look like in mind. We are thinking of using a Google sheet as a UI, and including a few mapping options for the user to pick from. Outputting several mappings would probably further increase our system's accuracy, because we wouldn't need the tool to pick the best mapping, but only to include it in the top k answers.

hanna-paasivirta:
Added the final experiments in this round of development. Details on the experiments can be found in the new experiment notebook focusing on the synonym issue `(synonym_mapping.ipynb`), and their background in the above comments.

The prompt versions are currently saved under experiments, and may be deleted in the next development phase. A summary of experiment results can be found in this [Google sheet](https://docs.google.com/spreadsheets/d/1oXmLQ1YWGVNGWjvn3IXzo8IhwV127hOoXhiEaT1lLDM/edit?usp=sharing). 

The next update will implement our current best synonym mapping pipeline out of the experiment notebook into a working Apollo service for our first usable demo.

hanna-paasivirta:
Nearly done with a v1 pipeline, a couple of small remaining things:
- Test with apollo and check input format
- Double check step outputs are exactly the same as before & evaluate a few times


TODOs longlist:
- Split prompts into other files like in other projects
- Move references to LOINC into the user input so the pipeline is more multipurpose. Simple change but needs testing to check the prompt change doesn’t drop the performance.
-  Add batching. Test what our API limits can reliably handle with this chain.
-  See how the keyword search could be more efficient
- The select 10 and select 1 steps both use the entire search results. This is more costly but testing suggested it gives better performance. Could revisit this if needs to be more efficient – check e.g. how often is the top 1 result the first result the model gives in top 10.
- Leave the topic of improving the performance for later, but one obvious quick tweak to try if there’s time is lowering the temperature closer to 0

josephjclark:
Great work on vocabmapper.py @hanna-paasivirta ! Lovely to see that in here. Let's work out how to get this live on Apollo on Monday

hanna-paasivirta:
Removed the LangChain implementation as it made debugging harder and served no purpose here over the Anthropic library. Now the mapper is tested and performance is the same as in experiments.

services/vocab_mapper/vocab_mapper.py
```diff
@@ -0,0 +1,159 @@
...
+    from dotenv import load_dotenv
+    from embeddings import loinc_store
+    from datasets import load_dataset
+
+    # Format Google Sheets data
+    input_data, column_indices, original_values = format_google_sheets_input(data)
+
+    # Initialize and setup mapper
+    load_dotenv(override=True)
+    ANTHROPIC_API_KEY = os.getenv("ANTHROPIC_API_KEY")
```
josephjclark, line 141: Note: we should read the api key from the incoming args object first, and if it's not there, default it from the env. Refer to `job_chat` for an example

josephjclark:
At the moment The vocab mapper server basically takes a spreadsheet in (a matrix of strings) and returns a spreadsheet. This is convenient for our demo integration in googlesheets and gets us to release way faster. So it's fine for now!

But later I'll want to refactor this to give a more formal (and easier to write) input and output structure, and move the data transformation to and from googlesheets into the Workflow.

josephjclark (approved):

PR #149, opened 2025-01-24
Job Chat: collections response update

### LINKED ISSUE #143: Job chat: update for collections
The AI Assistant is reporting mixed experiences with the collections API. 

Eventually we'll adopt a good RAG-based system which will add the right contextual docs. But until that's working, I want a quick-fix solution

We should address this urgently. Apparently `how does collections work` doesn't work, but `how to use collections in openfn` does

Image

Image

What is the minimum context we can add for collections? Can we just link people to the docs?

hunterachieng:
## Short Description

This PR focuses on giving the user the official documents on `openFn` docs

Fixes #143

## Implementation Details

Added a context to handle chat questions containing `collections`

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


services/job_chat/prompt.py
```diff
@@ -142,6 +142,12 @@ def generate_system_message(context_dict):
 
     if context.has("log"):
         message.append(f"<log>The user's last log output was :\n\n```{context.log}```</log>")
+    
+    if context.has("collections"):
```
josephjclark, line 146: I don't think this will work? This means "if the incoming payload has a `collections` key, append this to the prompt"

But Lightning doesn't send a collections key, so this will always be false. I suppose you could check to see if the word "collections" is in the user question - but that won't be very robust to typos, and it means the agent will never suggest a collections approach unless explicitly asked.

I also don't think Anthropic can access that information if you just pass it as a URL?

Have you tested this? What sort of response did you get?

hunterachieng: @josephjclark  I have changed the code and tested it locally. Here are the responses from two different prompts:
```
'Collections in OpenFn are a powerful feature that allow you to work with and manipulate sets of data more efficiently. They provide methods for querying, filtering, and transforming data within your jobs.\n\nFor working with collections, I recommend referring to the official documentation here: https://docs.openfn.org/adaptors/packages/collections-docs. \n\nThis documentation will provide you with detailed information on how to use collections effectively in your OpenFn jobs, including examples and best practices.'

```

```
'For working with collections, refer to the official documentation here: https://docs.openfn.org/adaptors/packages/collections-docs.\n\nThis documentation will provide you with detailed information on how to use collections in OpenFn, including examples and best practices.'
```

josephjclark:
Unfortunately I've fallen at the first hurdle here :(

image

Almost everything in that bullet list is incorrect. It's made it all up.

This is a follow up so the agent is probably being led on by its own history, but this is also nonsense:

image

This, in a new session, is fine. Fascinating how such a subtly different prompt gives a wildly different answer:

image

If I ask it something really specific, I also get a made up answer:

image

hunterachieng:
@josephjclark  I have added a notebook with a few examples and improved the prompt

josephjclark (approved):
Thank you Hunter - the adjusted prompt is much better, and the notebook really helps me easily understand the impact.

It's not perfect but I don't expect it to be. I am satisfied that this is an improvement.

I will hold off releasing to production until after the AI demo, but I'll close out the issue now.

PR #150, opened 2025-01-24
fix key retrieval for production env

hanna-paasivirta:
## Short Description

Fix the vocab mapper key retrieval for the production environment. Also add checks to see if all required keys can be found.

## Implementation Details

In the production environment, we get the keys from incoming requests or the system environment. This update checks for keys first in the incoming data, then the system environment. 

If the incoming data does not have an Anthropic key, then we can assume this is a development setting, and retrieve keys from an .env file.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/services/vocab_mapper/vocab_mapper.py
+++ b/services/vocab_mapper/vocab_mapper.py
@@ -6,7 +6,7 @@ import logging
 from vocab_mapper.prompts import *
 from vocab_mapper.tools import process_inputs
 from vocab_mapper.dataset_tools import format_google_sheets_input, format_google_sheets_output
-from util import create_logger
+from util import create_logger, ApolloError
 
 logger = create_logger("vocab_mapper")
 logging.getLogger('pinecone_plugin_interface.logging').setLevel(logging.WARNING)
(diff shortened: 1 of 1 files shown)


services/vocab_mapper/vocab_mapper.py
```diff
@@ -140,20 +140,37 @@ def main(data):
...
+
     # Format Google Sheets data
     input_data, column_indices, original_values = format_google_sheets_input(data)
     logger.info(f"Formatted input data: {input_data}")
 
-    # Get API key
-    if 'anthropic_key' in data:
-        ANTHROPIC_API_KEY = data['anthropic_key']
+    # Production: use data object keys -> system env
+    if 'anthropicApiKey' in data:
```
josephjclark, line 149: Instead of this if-else statement, can we do this?
```
load_dotenv(override=True)
ANTHROPIC_API_KEY = data.get('anthropicApiKey') or os.environ.get('ANTHROPIC_API_KEY')
OPENAI_API_KEY = data.get('openaiApiKey') or os.environ.get('OPENAI_API_KEY')
PINECONE_API_KEY = data.get('pineconeApiKey') or os.environ.get('PINECONE_API_KEY')
```

In other words we _always_ load dotenv, but we prefer keys passed in the data over keys defined in the environment,

services/vocab_mapper/vocab_mapper.py
```diff
@@ -140,20 +140,37 @@ def main(data):
...
+    missing_keys = []
+    if not ANTHROPIC_API_KEY:
+        missing_keys.append("ANTHROPIC_API_KEY")
+    if not OPENAI_API_KEY:
+        missing_keys.append("OPENAI_API_KEY") 
+    if not PINECONE_API_KEY:
+        missing_keys.append("PINECONE_API_KEY")
+
+    if missing_keys:
+        msg = f"Missing API keys: {', '.join(missing_keys)}"
```
josephjclark, line 171: This is great error reporting, thank you!

Can we raise an ApolloError, like job chat does? This will help ensure that the correct HTTP status is returned. And hopefully give us better log output too:

```
 raise ApolloError(500, f"Missing API keys: {', '.join(missing_keys)}", type="BAD_REQUEST")
```

services/vocab_mapper/vocab_mapper.py
hanna-paasivirta: Yes that would be simpler. With the if-else I was trying to make it easier to test the production scenario given the issue today, so that we don't accidentally read in an .env file when testing before launch. But I can try and make sure to delete it instead when testing instead.

josephjclark: I'll be setting up better ways to test this stuff next week, so we'll be able to worry about it a bit less. But the pattern a like this is what we'd usually do

PR #171, opened 2025-02-14
Add api key status check

### LINKED ISSUE #162: healthcheck: report API key status
When we call the healthcheck endpoint, we should report on the status of API keys. Whether they've been set and whether they're responding

We need a function that, on server start, calls some simple diagnostic API on any relevant services (at the time of writing it would be pinecone and openai). We need to find some simple no-op endpoint we can call just to validate the authentication

We should log the result on startup, maybe on the index page, and at the healthcheck endpoint

hanna-paasivirta:
## Short Description

Add a status check for API keys

Fixes #162 

## Implementation Details

Checks if Anthropic, OpenAI and Pinecone keys are working. It does not cost tokens to run the status check.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/platform/src/middleware/healthcheck.tsx
+++ b/platform/src/middleware/healthcheck.tsx
@@ -1,9 +1,19 @@
 import { Elysia } from "elysia";
 import pkg from "../../../package.json" assert { type: "json" };
+import { run } from '../bridge';
 
 export default async (app: Elysia) => {
   app.get("/livez", () => {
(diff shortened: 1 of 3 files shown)


josephjclark:
This looks great @hanna-paasivirta, thanks!

Next step is to try and call this service from the TypeScript code, using the bridge.

Maybe we can do this on a call on Monday?

josephjclark (approved):

Issue #173, opened 2025-02-17
Docsite search: Set update mechanism


hanna-paasivirta:
Once a basic docsite RAG service is done, design a mechanism to update the embeddings at appropriate intervals. Ideally, these would update only when the docs are updated, and only the parts that are updated would be regenerated, but this may be impractical.

josephjclark:
Yes - I think the solution here is that we do something like `POST services/embeddings/docs`, with credential/database data in the payload, which triggers the server to run the docs loader and update the database.

We can then trigger this from a github action whenever docs are merged to main.

I still want to think about what happens while the update is running. Do we drop the prod vector DB to update it? How long does the update take? How many RAG requests will fail because the database is updating?

hanna-paasivirta:
More ideas from Joe here: https://github.com/OpenFn/apollo/pull/176#discussion_r1981594195

I'm not sure how to increment parts of an online table though because it's chunked roughly by character length, and different types of changes might get complicated. It's a small dataset so regenerating should be ok.

josephjclark:
To investigate: does pinecone have any kind of versioning? How can we ensure the search is using the latest version of the embedded docsite?

josephjclark:
This is actually critical: we need to ensure we can organically update the embeddings every time the docsite is updated, and see the results in the assistant "immediately". So before we switch projects, we have to get this bit done.

josephjclark:
1. github action to update apollo embeddings when docsite is updated
2. work out how to get the latest collection

solutions?
- store the latest collection name in a separate table?
- ensure collection names are alphabetical (ie, `docs-${date}`), get all names, sort, and pick the top
- ?

PR #176, opened 2025-02-21
Docsite rag

### LINKED ISSUE #172: Docsite search: Add a new docsite search RAG
Add a new docsite search RAG to replace the experimental Search service. Use the Search service as an example, but rewrite to match the structure of our new embeddings services. 

- Leverage the Embeddings service and its SearchResult and VectorStore classes and add module to connect to docs embeddings
- Add separate service to embed docs
- Replace the Search service (doesn't seem to be in use) with a new docsite search service that can search by connecting to the Embeddings service 
- Use the same vector store and embeddings APIs as new services (Pinecone, OpenAI)
-  Analyse and optimise the chunking of the docsite and adaptor APIs
- Include metadata for each document returned to present to the user

hanna-paasivirta:
## Short Description

Replace the Search service with new embed_docsite and search_docsite services.

Fixes #172

## Implementation Details

The embed_docsite service downloads, chunks, processes metadata and indexes OpenFn documentation. The service uses Pinecone as a vector database and OpenAI for text embeddings. 

The search_docsite searches the documentation through the vector database using and input query.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI


hanna-paasivirta:
Todo: test on full docs; finish DocsiteSearch according to index metadata structure

josephjclark:
Thank you @hanna-paasivirta! I've gotta tied up with a production issue but we'll get this reviewed and merged tomorrow :)

services/embed_docsite/README.md
```diff
@@ -0,0 +1,38 @@
...
+To run directly from this repo (note that the server must be started):
+
+```bash
+bun py embed_docsite tmp/payload.json -O
+```
+
+## Implementation
+The service uses the DocsiteProcessor to download the documentation and chunk it into smaller parts. The DocsiteIndexer formats metadata, creates a new collection, embeds the chunked texts (OpenAI) and uploads them into the vector database (Pinecone).
+
+The chunked texts can be viewed in `tmp/split_sections`.
```
josephjclark, line 27: Ooh nice - this is still true? I'd love to run a test tomorrow and have a look, just to see what's going on

services/embed_docsite/embed_docsite.py
```diff
@@ -0,0 +1,57 @@
...
+    docsite_indexer = DocsiteIndexer(collection_name=COLLECTION_NAME)
+
+    # Add docs
+    for docs_type in DOCS_TO_UPLOAD:
+        # Download and process
+        docsite_processor = DocsiteProcessor(docs_type=docs_type)
+        documents, metadata_dict = docsite_processor.get_preprocessed_docs()
+
+        # Upload with metadata
+        idx = docsite_indexer.insert_documents(documents, metadata_dict)
```
josephjclark, line 54: So this doesn't clear the target collection first? It just adds new content?

Maybe we can add like a `purge` option or something? I'm gonna guess the default value is `true` but we can address that later

services/embed_docsite/github_utils.py
```diff
@@ -0,0 +1,101 @@
...
+        response = requests.get(url)
+        contents = response.json()
+        
+        # Handle single file response
+        if not isinstance(contents, list):
+            contents = [contents]
+
+        msg = contents[0].get("message") 
+        if msg and msg.startswith("API rate limit exceeded"):
+                logger.error("GitHub API limit exceeded")
```
josephjclark, line 48: Ah - when this happens do you remember how to fix it? Can you add a line of docs or a hint in this error?

Not critical, but it might make someone's day in the future

josephjclark (commented):
This is really nice and clean, thank you! Even I have a fighting chance of understanding it.

I'd like to test before merging but it looks fantastic. Left a couple of questions in the meantime.

services/embed_docsite/embed_docsite.py
hanna-paasivirta: It adds different doc types separately into the same collection (they can still be filtered by doc type), so it needs to not delete the contents. I think we might want to fill in a new collection for each version, then switch to that in production, then delete the old collection, rather than `delete_collection` and then refill/recreate it.

josephjclark: Right, we add docsite docs and adaptor docs separately to the same collection.

But the issue of knowing when to remove the old collection remains.

I like the idea of version switching. Incoming requests use the latest version, and once the version is updated, we can remove the old version (unless perhaps there are any outstanding requests, but at this stage the odds on a client making a call during an update feel pretty slim).

I think this PR I'd like to:
- have some kind of version tracking in the db (maybe we need a meta table for this)
- Whenever there's an update, lookup the current version and increment it
- Insert all new docs into the current version
- Update the meta table
- Drop the old version

We can also lookup the  version number when the server starts up, store it in-memory for incoming requests, and update it in-memory whenever there's an update.

I suggest we open up a new issue for this and implement a solution on a new branch. It's not crazy complicated, but probably risky enough that we should make the change in isolation. I can see a few places where it might go wrong 😅

josephjclark:
Is `embed_docsite/search.py` still used? That looks like the old stuff?

services/embed_docsite/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Payload Reference
+The input payload is a JSON object with the following structure:
+
+```js
+{
+    "docs_to_upload": ["adaptor_docs", "general_docs", "adaptor_functions"], // Select 1-3 types of documentation to upload
```
josephjclark, line 34: I think we should be able to default all this

We should generate the collection name according to the versioning strategy (see my earlier comment). To be fair, versioning by date is also a great solution and I'm quite happy to roll with it.

By default we should upload all docs, but users should be able to specify fewer if they want.

It would be more conventional to ask the user to pass an api key and pinecone URL. Maybe apollo should have its own credentials for this.. but also maybe not? I think we should give this more thought.

But users must be able to pass credentials, even if if apollo provides a default

josephjclark:
@hanna-paasivirta what do I need to setup on the pinecone side to get this working? I'm getting an error because I don't have a `docsite` index. When I go to create an index, I get all sorts of complicated questions about how I should configure it :grimacing:

services/embed_docsite/embed_docsite.py
hanna-paasivirta: Yep, continued on here https://github.com/OpenFn/apollo/issues/173

hanna-paasivirta:
Index creation should work properly now.

Next up figuring out how to actually make use of the RAG https://github.com/OpenFn/apollo/issues/180

josephjclark:
This is working great now, thank you @hanna-paasivirta ! I'm just having a little look over the chunks and implementation

josephjclark:
Can we add some facility to exclude some paths from indexing?

For example, the roadmap or `job-examples.md` feel like they're in danger of including misleading information. Release notes probably  aren't helpful (what if it randomly decides an old version of salesforce has the most relevant informaton?).

I'd rather not index this stuff at all. I don't know how harmful these docs could be but I feel like a general utility to exclude stuff would be useful.

I'd suggest:
1) a default, very conservative, exclude list is baked into the processor (or indexer mayber?(
2) clients can submit their own exclude list

services/embed_docsite/docsite_indexer.py
```diff
@@ -0,0 +1,142 @@
...
+        docs = loader.load()
+        logger.info(f"Inputs processed into LangChain docs")
+        logger.info(f"Uploading {len(docs)} documents to index...")
+
+        idx = self.vectorstore.add_documents(
+            documents=docs
+        )
+        sleep_time = 30
+        logger.info(f"Waiting for {sleep_time}s to verify upload count")
+        time.sleep(sleep_time)
```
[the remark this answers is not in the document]
hanna-paasivirta: I added a while loop and a max wait time instead. This works in the current implementation where we’re not expecting several indexing jobs running in parallel.

services/embed_docsite/github_utils.py
```diff
@@ -0,0 +1,102 @@
...
+    # Download and process each file
+    files_data = download_main_docs(github_files)
+            
+    return files_data
+
+def get_docs(docs_type):
+    if docs_type == "adaptor_functions":
+        return get_adaptor_function_docs()
+    if docs_type == "general_docs":
+        return get_github_path_contents(repo="docs", path="docs")
```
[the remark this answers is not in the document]
hanna-paasivirta: Yes, but it won't download the history now so I thought it might be more lightweight? I didn't test it though

services/embed_docsite/docsite_processor.py
```diff
@@ -0,0 +1,187 @@
...
+
+    :param docs_type: Type of documentation being processed ("adaptor_functions", "general_docs", "adaptor_docs")
+    :param output_dir: Directory to store processed chunks (default: "./tmp/split_sections").
+    """
+    def __init__(self, docs_type, output_dir="./tmp/split_sections"):
+        self.output_dir = output_dir
+        self.docs_type = docs_type
+        self.metadata_dict = None
+
+    def _write_chunks_to_file(self, chunks, file_name):
```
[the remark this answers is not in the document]
hanna-paasivirta: - I moved public functions to the top and shortened unnecessary doc strings 
- I’m a bit hesitant to remove all underscored functions because I rely on them quite a bit. I don't use them all the time, but some of these classes like docsite_processor have lots of similar sounding functions that are designed to be run as part of a strict and fiddly data processing pipeline. I wouldn’t be able to remember which one(s) I’m supposed to run, and would have to write more instructions in the doc strings that would take longer to understand than glancing at the underscores. In general, I think they’re also used e.g. when doing import *, to avoid importing them, and sometimes when processing the code itself, e.g. in linters, or LLMs e.g. “check the doc strings in the internal functions in this repo are consistently one line only”. What do you think?

services/embed_docsite/docsite_processor.py
```diff
@@ -0,0 +1,187 @@
...
+
+        return text.strip()
+
+    def _split_by_headers(self, text):
+        """Split text into chunks based on Markdown headers (# and ##) and code blocks."""
+        sections = re.split(r'(?=^#+\s.*$|^```(?:.*\n[\s\S]*?^```))', text, flags=re.MULTILINE)
+
+        return [chunk.strip() for chunk in sections if chunk.strip()]
+
+    def _split_oversized_chunks(self, chunks, target_length):
```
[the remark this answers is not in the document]
hanna-paasivirta: Yes it is, but it's important for chunks to be a similar size if we're using semantic search. I put this in a separate issue we can tackle https://github.com/OpenFn/apollo/issues/181 . Generally, it's probably not that useful to spend time on chunking logic (vs. search logic, data augmentation), because it's time consuming and doesn't necessarily improve results when we're fetching multiple chunks per search. It's quite standard to just split text by character counts. If we move away from semantic search though (e.g. search by title/summary https://github.com/OpenFn/apollo/issues/180) aiming for self-contained sections might become more important though.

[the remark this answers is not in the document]
hanna-paasivirta: At this stage where we're splitting oversized chunks, we've already split into headers and code blocks. The oversized chunk splitting runs after and references the max chunk size setting.

services/embed_docsite/embed_docsite.py
```diff
@@ -0,0 +1,77 @@
...
+        logger.error(msg)
+        raise ApolloError(500, f'Missing API keys: {", ".join(missing_keys)}. Add to payload or environment.', type="BAD_REQUEST")
+
+    # Initialize indexer
+    docsite_indexer = DocsiteIndexer(collection_name=collection_name, **other_params)
+
+    # Add docs
+    for docs_type in docs_to_upload:
+        # Download and process
+        docsite_processor = DocsiteProcessor(docs_type=docs_type)
```
josephjclark, line 70: Sorry, just before we merge and forget about this, can we feed docs_to_ignore  into the processor from here from user args?

And can we then update the readme with:
a) one example with default values (which will do everything)
b) one example which sets docs_to_upload and _docs_to_ignore

services/embed_docsite/github_utils.py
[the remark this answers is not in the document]
josephjclark: I've raised #182 as something to pick up on a slow friday afternoon

services/embed_docsite/docsite_processor.py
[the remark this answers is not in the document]
josephjclark: Got it: if you rely on the underscore convention, that's reason enough to keep them. Thanks!

services/embed_docsite/embed_docsite.py
hanna-paasivirta: Sorry I forgot to commit this file, I've added it in now along with the readme.

josephjclark (approved):

PR #197, opened 2025-03-25
update docsite embeddings automatically

### LINKED ISSUE #173: Docsite search: Set update mechanism

hanna-paasivirta:
## Short Description
Add
a) github action to update apollo embeddings when docsite is updated
b) Add a step to enable using the latest collection by default

Fixes #173

## Implementation Details

a) Adjust docsite upload functionalities for automated updates (to be set in GitHub actions)
- change docsite naming convention to have more precise time (YYYYMMDDhhmm) in case there are multiple docsite udpates in a day
- add max_total_collections parameter to delete oldest collection if there's more than n collections with the name format "docsite-YYYYMMDD" (ignore other collections e.g. "docsite-test")

b) Accessing the latest collection:
Adds a step in the Search Docsite service to get the names of collections within an index, and use the collection with the latest date, unless a specific collection is not requested. This is done in  `_get_most_recent_namespace` in `DocsiteSearch`. 
When called from the AI assistant, this would run every time we query the database (once at the beginning of a conversation).

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/services/embed_docsite/README.md
+++ b/services/embed_docsite/README.md
@@ -27,21 +27,14 @@ The service uses the DocsiteProcessor to download the documentation and chunk it
 The chunked texts can be viewed in `tmp/split_sections`.
 
 ## Payload Reference
-The input payload is a JSON object. A simple example with default values:
-
-```js
-{
-    "collection_name": "Docsite-20250225", // Name of the collection in the vector database.
-}
-```
-
-An example with all parameters:
+The input payload is a JSON object. All parameters are optional:
 
 ```js
 {
     "docs_to_upload": ["adaptor_docs", "general_docs", "adaptor_functions"], // Select from 3 types of documentation to upload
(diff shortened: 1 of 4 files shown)


josephjclark:
TODO: the POST/trigger endpoint on apollo should insist on api keys being included in the payload (and those api keys should be used in place of the environment). This just ensures that the work is being paid for properly.

Return a 403 if credentials are not included.

We should also probably validate the credentials using that status check we have. Let's not be fooled by `OPENAI_API_KEY=abc`

hanna-paasivirta:
The code here is ready to merge with the other RAG updates, but the GitHub action still needs setting up to trigger the creation of new docsite vectors.

PR #199, opened 2025-03-28
Assistant docsite rag

hanna-paasivirta:
## Short Description

Add v1 docsite RAG for the AI assistant.

Fixes #

## Implementation Details

Add a docsite semantic search at the beginning of an AI assistant conversation. 

Search trigger & process
This search uses the user's first message (and adaptor info as context) to decide whether to search our documentation. If so, it will generate one or more queries (with optional adaptor only/general docs only filters). These are used to perform a semantic search with a threshold & max docs limit. The returned documents are injected in the prompt.

Versioning, monitoring
- Automatically gets the latest docsite
- Returns doc retrieval info

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI


hanna-paasivirta:
Todo:
- merge other PRs into this (claude version, docsite updates) 
- Check when prompt generation cadence
- Check if search results formatting is optimal
- Test run & performance, possibly adjust threshold/max doc slightly

josephjclark:
## What does rag cost?

We need to know the total tokens consumed by

rag search
non rag search

This tells us how much rag affects billing

## When to run rag?

If the incoming data_dict does NOT have a `prompt` key, we run rag. If it DOES have a prompt key, we skip rag and pass the prompt straight through

Later (maybe very soon!) we'll refine and improve this

## What are we actually billing for?

I need to check lightning/thunderbolt and check what usage is currently being billed for. And from there, we work out how rag tokens affect this

josephjclark:
Oh, also: if we remove the static adaptor docs (which must be like 5k tokens), what difference will that make to the overall token usage?

It would be great if the extra tokens of the RAG are offset by the existing bloat of the adaptor doc stuff

josephjclark:
Left to do before release:
- Adding data to lightning
- Maybe function docs thing ?
- Automatically update vectors when docsite is updated (or, support it manually and we'll add the action later)
- QA

hanna-paasivirta:
The GitHub action for automated docsite updates still needs setting. It should run the embed_docsite service with just the following input:

`{
    "max_total_collections":2
}`

josephjclark:
Ah, one large error has just come up because the rag has returned hits from multiple adaptors, and so the assistant is confused about which API  is available. So it's included code from the wrong adaptor.

josephjclark:
I am pretty sure that we should wait and fix the functions thing before releasing this. I don't think it's huge amounts of work (not for a first pass), and removing the d.ts download feels super important to me.

hanna-paasivirta:
> Ah, one large error has just come up because the rag has returned hits from multiple adaptors, and so the assistant is confused about which API is available. So it's included code from the wrong adaptor.

I could add a contextualising sentence for each retrieved chunk e.g.  "from <title> docs:  <chunk>" ? Theoretically the user could be asking about other adaptors so we probably don't want to filter by section title for the adaptors?

PR #202, opened 2025-04-02
Assistant questions

### LINKED ISSUE #154: job chat: create a series of test questions
This is basically an integration test suite

We should come up with a list of questions, run them against the assistant, and MANUALLY review the answers.

This is the sort of thing we might do ahead of every release.

Maybe, with some sort of automation magic, we can build them into an integration test

The previous answer needs to be commited to the repo, so that we can manually compare the diff.

Questions:

* How do I use collections?
* How do I get a list of items in my collection created after a set date?
* How do I get and update a tracked entity in DHIS2?
* What is state and how do I use it?
* What is an Operation?
* [http] I've fetched a resource from my endpoint but I don't know how to get hold of it?
* How do I get permission/authorisation to update my googlesheet?

josephjclark:
## Short Description

This PR sets up static list of [14] questions, sends them to the assistant, and writes the result to disk

Fixes #154

## Implementation Details

I've written this in typescript because it's quick and easy for me - but it does mean I have to have the server running, and I've harded coded URLs in to the code. It's not great.

We should convert it to python and call into the job_chat module directly. Way easier. Then we can add some docs about how to run it.

The questions are pulled together fairly quickly and I've only lightly reviewed the quality of answers. But it doesn't matter how they perform on main  - what I'm interested in is how they change when we test against the RAG branch.

The results are written to disk so when we look at a PR on github, we should get an idea of what's changed.

I know the results aren't deterministic and there will be variation even on main. But having these tests (or something like them) will help me understand the impact of changes we're making.

Next step might be to run the question multiple times, and use AI to collate the answers and flag any meaningful differences? Difficulty being that "meaningful" is very subjective. But I think that's a challenge for another day.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/services/job_chat/evaluation/load-questions.ts
+++ b/services/job_chat/evaluation/load-questions.ts
@@ -1,0 +1,51 @@
+export type Question = {
+  question?: string;
+  adaptor?: string;
+  code?: string;
+  input?: string;
+};
+
+export const loadQuestions = async (
+  path = "questions.md"
+): Promise<Question[]> => {
+  const foo = Bun.file(path);
+
+  const questionsContent = await foo.text(); // contents as a string
+
+  const lines = questionsContent.split("\n").filter((l) => l.length);
+
+  const questions = [];
+
+  let q: Question = {};
+  let content: string[] = [];
+  let key: keyof Question = "";
+  while (lines.length) {
+    const next = lines.shift();
+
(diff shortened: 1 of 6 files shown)


josephjclark:
To run, go into the folder with the .ts stuff and run `bun test.ts`

Issue #213, opened 2025-04-25
Create new service to generate project yaml


hanna-paasivirta:
Goal: 
- we'll add a workflow_chat service
- it's designed to iterate/edit workflows (as yaml)
- it doesn't accept or return the body key on steps

KR1: Have a working service to replace the current `gen_project`. Use a similar structure to `job_chat` but return a) generated text b) a yaml. 
- Needs a single baseline prompt that can flexibly produce a request for clarification OR a yaml+explanation.
- The model needs to process the conversation history + the prompt
- The answer will be split into text & yaml (if present) on a rule-based basis
- Empty body key generated by LLM (can be replaced elsewhere)
- Think of how to best return yaml inside the json that is sent to Lightning
Timeline: ready for **1 May**

KR2: Experiments to improve quality of responses from the baseline solution
- Gather more complex yaml examples from the team
- exp 1: Split the baseline prompt logic into two or more discuss & generate yaml calls (+ decision to generate) inside the workflow chat service so that each task can be improved separately.
- exp 2: improve the selected llm call logic through prompt-tuning the baseline prompts. Use company documentation e.g. sales script to guide the discussion prompt.
- + address other issues surfaced after kr1
Timeline: **7 May**?

KR3: Work with the team to develop the service
- Work with Aisha/Aleksa to develop the prompts (what info to ask for, when to generate)
Timeline: **13 May**?

josephjclark:
3 issues to explore going forwards:

1. Adding extra information to the prompt for better workflows
2. Externalising some prompt content outside of apollo for sales team to customise
3. Allowing proper, responsible edit of an existing workflow (preserving/rerwiting job code as appropriate)

josephjclark:
@hanna-paasivirta when you have time this week, could you make sure that everything in this issue that hasn't been done, including my comments above, is spun out into its own issue? Then we can close this 👍

PR #214, opened 2025-04-29
Add service to generate workflow.yaml

hanna-paasivirta:
## Short Description

Replace the current `gen_project` with `workflow_chat`, a new LLM-based service that can chat with a user with the goal of generating a workflow.yaml. It can be called with a conversation history and current YAML to generate a new workflow.yaml.

See #213 for plan

## Implementation Details
This is a working first version for the workflow chat service for dev purposes. The internal structure and prompts will still be further developed before launching to users.

The service uses a single LLM prompt that can flexibly produce a request for clarification OR a yaml+explanation.  The prompt has information about workflows, YAML structure, adaptors. We can also give the model the conversation history and existing YAML.

The inputs are `content` the user's conversation turn, `history` conversation history (optional), `existing_yaml` an existing YAML (optional).

The service returns `response` text (LLM conversation turn), `response_yaml` a YAML if generated and model `usage`. The body key is currently generated by the model, and should by default be a commented out, empty field.

See the README for examples.

Some of the remaining issues are: unreliable splitting of the YAML from the model response; dealing with the body key safely either in Lightning or in Apollo (if we allow it to have any content when using this service).

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI


hanna-paasivirta:
@elias-ba I try and deal with the YAML formatting here: https://github.com/OpenFn/apollo/blob/yaml-gen/services/workflow_chat/workflow_chat.py#L113 But I'm not sure if this is enough to make it functional for you.

josephjclark:
This is ready for release subject to some basic QA from Elias

services/workflow_chat/gen_project_prompts.yaml
```diff
@@ -0,0 +1,190 @@
...
+        condition_type: always
+        enabled: true
+      job-one->job-two:
+        source_job: job-one
+        target_job: job-two
+        condition_type: on_job_success
+        enabled: true
+    ```
+
+    ## Adaptor Knowledge
```
josephjclark, line 59: We need to spin out an issue for a better solution to encoding this adaptor list.

It feels like knowledge of what openfn adaptors are available should be a standard, generated and pre-packaged thing that all apollo services can use? A nice bit of infrastructure that can be re-used in all prompts.

I don't know how we'll do it in practice. Maybe apollo could fetch the list when it loads. Or we treat it like embeddings, and add the knowledge to a database and update automatically it whenever there's an adaptor release.

Anyway, first step is to track the work in an issue.

services/workflow_chat/workflow_chat.py
```diff
@@ -0,0 +1,172 @@
...
+
+        response_parts = []
+        for content_block in message.content:
+            if content_block.type == "text":
+                response_parts.append(content_block.text)
+            else:
+                logger.warning(f"Unhandled content type: {content_block.type}")
+
+        response = "\n\n".join(response_parts)
+        response_text, response_yaml = self.split_format_yaml(response)
```
josephjclark, line 92: Maybe this is an issue for later, or maybe the model is just good enough, so this is an open question

Should we run some validation on the yaml to ensure that all suggested adaptors and versions actually exist? And if not, go back to the model?

Maybe a good first step would be to detect unrecognised adaptors and log it somewhere. Than we can use that as a metric to determine if we need to do more work here.

services/workflow_chat/README.md
```diff
@@ -0,0 +1,58 @@
+# workflow_chat Service
```
josephjclark, line 1: Add a note about how error handling works right now

services/workflow_chat/workflow_chat.py
hanna-paasivirta: I added a logger warning for now. Do you have a recommended way of sending a slack notification or other logging method?

josephjclark: Let's raise an issue to send an error back with the payload and warn users that maybe the data isn't quite right

josephjclark (approved):

PR #258, opened 2025-07-09
workflow_gen: Add UUID preservation to workflow generation

### LINKED ISSUE #257: workflow_chat: Workflow generation for existing workflows leads to UUID re-generation
The current implementation of workflow generation (allowing existing workflow edits) only outputs a YAML. However, Lightning uses JSON Patches to adjust workflows, and in this process, it changes all the existing UUIDs. This could lead to subtle problems in audit trails that the user would not be aware of.

In order to preserve UUIDs in an existing workflow, Lightning will start to send these as part of the workflow YAML. Apollo should aim to preserve these and insert new UUIDs for new workflow steps and edges.

hanna-paasivirta:
## Short Description

This PR helps preserve UUIDs when editing existing workflows with the workflow chat service. It can preserve existing UUIDs and generate new UUIDs for new jobs/edges. 

The service cannot subtly edit existing UUIDs, but it could accidentally re-generate them.

Fixes #257 

See Lightning work: https://github.com/OpenFn/lightning/issues/3247

## Implementation Details
The input JSON format is not changed, but the "existing_yaml" key now includes "id" keys under triggers, jobs and edges.

The UUID preservation follows the existing approach we use for job code preservation: we replace the UUIDs with placeholder strings, that the model is less likely to forget. Then we send the YAML without UUIDs (and without job code) to the LLM. We expect the model to preserve the placeholder strings in the appropriate places (or delete them where necessary) in its output YAML. We then replace these strings with the original UUIDs using the placeholder strings. Where a replacement is not found (if there is a new job/edge, or if the model forgot the correct placeholder string), we generate a new UUID.

We use curly braces and job/edge-name-based placeholder strings to help the LLM remember where UUIDs belong, and to differentiate from code placeholder strings, which use underscores.

Job and edge ID placeholder strings are sent to the model. There can only be one trigger per workflow, so this is not sent to the model, but preserved and added in automatically. Changes to jobs/triggers should not usually lead to a new UUID.

The prompt is adjusted to explain the purpose of the ID placeholders and to ask the model to preserve them where appropriate. The model is instructed to leave ID fields empty for new sections. Whether they are empty or if the model invented a new placeholder string (against instructions), these will be filled in in `workflow_chat.py`

Tests are adjusted for the new YAML format. (TODO)

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.changeset/smart-rats-shake.md
+++ b/.changeset/smart-rats-shake.md
@@ -1,0 +1,5 @@
+---
+"apollo": patch
+---
+
+workflow_chat: preserve UUIDs during workflow generation
--- a/services/workflow_chat/gen_project_prompts.yaml
+++ b/services/workflow_chat/gen_project_prompts.yaml
@@ -22,10 +22,12 @@ prompts:
     name: open-project
     jobs:
       job-one:
+        id: {{ID_JOB_job-one}}
         name: First Job
         adaptor: "@openfn/language-common@latest"
         body: "// Add operations here"
       job-two:
+        id: {{ID_JOB_job-two}}
         name: Second Job
         adaptor: "@openfn/language-http@latest"
         body: "// Add operations here"
@@ -41,11 +43,13 @@ prompts:
         enabled: false
     edges:
       daily-trigger->job-one:
+        id: {{ID_EDGE_daily-trigger->job-one}}
         source_trigger: daily-trigger
         target_job: job-one
         condition_type: always
         enabled: true
       job-one->job-two:
+        id: {{ID_EDGE_job-one->job-two}}
         source_job: job-one
         target_job: job-two
         condition_type: on_job_success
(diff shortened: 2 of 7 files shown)


hanna-paasivirta:
> Hey @hanna-paasivirta, I just found out that Apollo would return the placeholder ID (like `{ID_JOB_Send-Email}` or `{ID_EDGE_webhook->Send-Email}`) when the user requests a rename on nodes or edges. This was happening because the restore_components function was looking up IDs by the new component name (example `job_id_Notify-User`) but the preserved_values dict only had the old name (example `job_id_Send-Email`). I just pushed a fix [here](https://github.com/OpenFn/apollo/pull/258/commits/1562bbcdb2ca54b1f398d3d36f4a8348dd631647) using a placeholder lookup approach - it builds a reverse map from placeholder strings to IDs and matches them directly, making it completely rename-proof. Let's talk about it tomorrow morning. I am happy to revert my commit if you think we could do better but for now this fixes the problem very well.

Thank you so much @elias-ba , you're absolutely right, I had implemented the restoration step wrong.

I have kept your fix, but gone back to using a single preservation dict & removed the nested structure so that we don't have to create the reversed dict.

I added some tests to check for body and id keys, and (just one) test for renaming jobs. Please let me know how it looks on your side now.

josephjclark (approved):

PR #259, opened 2025-07-16
Make job_chat return edited code

### LINKED ISSUE #256: job_chat: attach replacement code
We want to improve the UX of Lightning's AI assistant so that code suggested by the model can be automatically applied to user's code.

We will achieve this by:
- Returning a key like `suggested_expression` in the final payload
- Returning a key like `suggested_diff` in the final payload
- Have the model return a diff and explanation, rather than inline code snippets in the chat
- Apply the diff suggested to the model to the original text and send it back to the client
- Also include the raw diff that goes back to the client

Additionally, we will impose a limit (5000 characters?) with a default in the code and an override in an env var. This constraints the size of the expression that we will send to the model.

We should aim to make changes which break Lightning as little as possible.

Note that Lightning may send a partial expression (ie just the selected code) rather than the whole job. This is fine. Apollo won't know and should assume that what it is sent is the whole expression.

hanna-paasivirta:
## Short Description

Gives the AI Assistant (job_chat) the ability to suggest edits to a user's job code.

The service can now return a suggested_code key if relevant. This key contains the complete code for the job, with changes applied to it by an LLM.

The input payload remains unchanged; the job code is contained in `"expression".` 

Fixes #256
Introduces #262 

## Implementation Details

Internally, the LLM outputs edits to specific sections of the code, and these are applied to the full code before returning it to Lightning

The service introduces:
- An updated prompt for the LLM
- LLM returns JSON with `text_answer` and `code_edits` keys
- Replace/rewrite actions only 
- Edits/inserts are applied with string replacements
- The prompt asks the LLM to avoid duplicate code problems/string replacement errors by telling the model to replace larger/unique sections. 
- There is a separate correction LLM call that is triggered if the edits cannot be applied to the job code. This can run only once (to avoid an infinite loop). This is currently not summed into the total token cost.
- Tests for both the old and new service structures are added in`/tests`

The service is backwards compatible and the old prompt which elicits only a string response from the LLM (with embedded code) can be used instead of the new prompt that can output code edits.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI


josephjclark:
Let's raise a little issue to track the (hopefully very rare!) case where the model will lose count and replace the wrong bit of code. A reliable repro case in the issue would be useful. But we won't invest time in it until we get feedback from users

services/job_chat/job_chat.py
```diff
@@ -140,9 +300,25 @@ def main(data_dict: dict) -> dict:
...
+            "response": result.response,
+            "suggested_code": result.suggested_code,
+            "history": result.history, 
+            "usage": result.usage, 
+            "meta": {"rag": result.rag}
+        }
 
-        return {"response": result.content, "history": result.history, "usage": result.usage, "meta": {"rag": result.rag}}
+        if result.application_status:
+            response_dict["application_status"] = result.application_status
```
josephjclark, line 319: Joe to review

josephjclark:
Need to be careful about how we release this: this will break code editing for existing users until Lightning has been updated

services/job_chat/tests/test_pass_fail.py
```diff
@@ -0,0 +1,449 @@
...
+  return { ...state, profile };
+});
+
+// Save the data
+post('https://api.example.com/save', state => ({
+  user: state.profile,
+  timestamp: new Date().toISOString()
+}));'''
+
+    expected_code_alternative = '''
```
josephjclark, line 124: I'm being a bit fussy here but this "alternative" pattern isn't so nice.

The problem I guess is that sometimes the model, quite reasonably, will add an empty object to the  request.

If we changed this to `replace GET to DELETE` - or maybe the other way around - we'd probably get a cleaner test

You can also to PUT to POST and always include the argument, like `put('www', { body: $.data })`

services/job_chat/tests/test_pass_fail.py
```diff
@@ -0,0 +1,449 @@
...
+  state => ({
+    source: 'Patient API',
+    results: state.patients,
+    summary: state.summary,
+    endpoint: getPatientData('summary')
+  })
+);'''
+    }
+    
+    expected_code = '''// Custom helper functions
```
josephjclark, line 237: Presumably, rather than duplicating all this code, you could just do a simple string replace on the function name? That also has the benefit of being easier to read for humans. Because I, as a test reviewer, can't easily tell you the difference between the original and expected code snippets

services/job_chat/tests/test_pass_fail.py
```diff
@@ -0,0 +1,449 @@
...
+    assert "suggested_code" in response
+    
+    assert response["suggested_code"] == expected_code, f"Function rename did not produce expected result.\nExpected:\n{expected_code}\n\nActual:\n{response['suggested_code']}"
+
+def test_change_multiple_instances():
+    print("==================TEST==================")
+    print("Description: Testing changing multiple instances of a URL path segment across longer job code.")
+    
+    history = []
+    content = "change all the occurences of endpoint to endpoooint"
```
josephjclark, line 331: Doing this  the other way around, `endpoooint -> endpoint`, would feel like a more realistic usecase - using the model to fix an error.

What would be the consequences of having `endpoooint` in a string and expecting the model to ignore it? Like "rename all  variables" rather than "replace all values" is a more interesting and realistic AI test.

But a) maybe that's testing the wrong thing, and b) it introduces more scope for variation. What do you think?

services/job_chat/tests/test_qualitative_old_prompt.py
```diff
@@ -0,0 +1,194 @@
+import pytest
+import json
+from .test_utils import call_job_chat_service, make_service_input, print_response_details
+
+
+def test_basic_input():
```
josephjclark, line 6: Are the qualitative tests for the old prompt of any value? I think a couple of basic pass/fail, just to ensure the prompt is loaded at the right time, feels like enough. Because right now we're basically quality testing legacy code, which seems of little value

services/job_chat/tests/test_qualitative.py
```diff
@@ -0,0 +1,201 @@
+import pytest
+import json
+from .test_utils import call_job_chat_service, make_service_input, print_response_details
+
+
+def test_basic_input():
```
josephjclark, line 6: I'm sure we've talked about this but I can't  remember the outcome :blush:  Do we want to commit these results into github so that we can a) more easily assess quality and b) more easily understand the impact of changes over time?

josephjclark (commented):
Hi @hanna-paasivirta, left a few thoughts on comments on this. Take a look and we can maybe discuss tomorrow!

services/job_chat/tests/test_qualitative.py
hanna-paasivirta: We could commit them, but looking at one output for one version has no statistical significance and could be misleading. These are basic tests focusing on functionality rather than actual model evaluations datasets where we might eventually use a more LLM-focused library and somewhere other than GitHub to store datasets.

PR #264, opened 2025-07-30
Truncate adaptor strings in job_chat prompts

### LINKED ISSUE #263: job_chat: adaptor docs from apollo("describe_adaptor") are too long
**Problem**: Users can hit token limits in a fresh conversation with the AI Assistant, even with code/logs/history.

Error reported by Aleksa:
1. `Error: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'prompt is too long: 204137 tokens > 200000 maximum'}}`
2. My job code is only 31 lines -> not every long. Note screenshot where I am only checking code box
3. The question I’m asking is pretty short: 
what does this job error mean? Run complete with status: fail JobError: getaddrinfo ENOTFOUND kf.kobotoolbox.orghttps
… why am I hitting token limits so easily? 

**Cause**: The long Adaptor docs for the http adaptor version 3.1.11. 

I checked it like this:
```
from apollo.services.util import apollo
adaptor = "@openfn/language-http@7.2.0" 
adaptor_string = "" 
adaptor_docs = apollo("describe_adaptor", {"adaptor": adaptor}) 
for doc in adaptor_docs: 
    adaptor_string += f"Typescript definitions for doc {doc}" 
    adaptor_string += adaptor_docs[doc]["description"] 
adaptor_string += "</adaptor>" 
print(len(adaptor_string))
% 807624
```
For the latest version 7.2.0 the character length is 34853, so this fits in the prompt.

**Solution**
- (Quick fix will be merged, with truncation)
- First check why the descriptions can be so long. Is it a mistake? Are there more relevant parts we could use instead?
- We originally didn't go for RAG for this part (unlike general docs) because this seems hard to search with just similarity calculation. Could verify if this is really the case. 
- If we can't filter these down in some deterministic way, and RAG is as bad as expected, find another solution. (Separate LLM call to select a section of these function docs?)

hanna-paasivirta:
## Short Description

Truncate adaptor description strings to 35k characters. 

Fixes #263

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.changeset/proud-gifts-grow.md
+++ b/.changeset/proud-gifts-grow.md
@@ -1,0 +1,5 @@
+---
+"apollo": patch
+---
+
+job_chat: truncate adaptor docs to 40k characters
--- a/services/job_chat/prompt.py
+++ b/services/job_chat/prompt.py
@@ -168,6 +168,9 @@ def generate_system_message(context_dict, search_results):
         for doc in adaptor_docs:
             adaptor_string += f"Typescript definitions for doc {doc}"
             adaptor_string += adaptor_docs[doc]["description"]
+        if len(adaptor_string) >= 35000:
+            adaptor_string = adaptor_string[:35000]
+            adaptor_string += "(...)"
         adaptor_string += "</adaptor>"
 
         message.append(adaptor_string)
--- a/services/job_chat/prompt.py
+++ b/services/job_chat/prompt.py
@@ -168,8 +168,8 @@ def generate_system_message(context_dict, search_results):
         for doc in adaptor_docs:
             adaptor_string += f"Typescript definitions for doc {doc}"
             adaptor_string += adaptor_docs[doc]["description"]
-        if len(adaptor_string) >= 35000:
-            adaptor_string = adaptor_string[:35000]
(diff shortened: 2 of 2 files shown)


josephjclark:
It looks like common - one of our biggest - is 23k characters. We include that in every request. So let's say 25k for common.

Then you've got another adaptor on top: commcare is 8.5, salesforce is 17.5k, dhis2 is 20k :scream: 

So if we're going to rush a fix in for this, I think we probably have to go as far as a 50k character limit

I'd really love to get on soon and improve this whole thing: finding a way to reduce the quanity of adaptor docs, to only select the relevant stuff, and to avoid re-downloading on every request. Might have to accelerate to that soon.

hanna-paasivirta:
@josephjclark Yes I thought we might push a quick fix as this is blocking AI assistant usage on a project Aleksa is working on, then schedule time to fix it properly.

I didn't have a look at all the versions, but for @latest, 35k characters is about 80th percentile, 45k is 95th percentile, longest is 643k.

josephjclark:
Let's just say 40k for the moment - thanks!

PR #276, opened 2025-09-03
workflow_chat: add job name sanitisation for edges

### LINKED ISSUE #277: workflow_chat: The LLM can add a trailing single quotation in YAML
Error reported by Aisha:

> Getting an error when asking the ai assistant(workflow builder) to remove the first node of the workflow and to change the adaptor for the last workflow. I tried to separate the requests into two but still getting an error:
**Job reference error: The job 'Sync-to-DHIS2'' referenced in edge 'Sync-to-DHIS2->Send-Error-Report-to-Admins' does not exist. Please check that all job names are correctly spelled and hyphenated.**

There is an extra quotation mark in the job name 'Sync-to-DHIS2'' (thanks Joe!). When testing, we can see that the model adds this at the end of the YAML by accident as it loses track of quotation marks.

Our current YAML sanitisation doesn't sanitise job names inside edges. Expanding the sanitisation rules will fix these kinds of errors (although not the root cause).

hanna-paasivirta:
## Short Description

Expands YAML sanitisation from job names to edges (edge key, source job, target job)

Expands testing to check this sanitisation is done for all outputs, including edges.

Fixes #277

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.changeset/clean-fishes-guess.md
+++ b/.changeset/clean-fishes-guess.md
@@ -1,0 +1,5 @@
+---
+"apollo": patch
+---
+
+add sanitisation to edges in yaml
--- a/services/workflow_chat/tests/test_pass_fail.py
+++ b/services/workflow_chat/tests/test_pass_fail.py
@@ -7,7 +7,7 @@ import subprocess
 from pathlib import Path
 import difflib
 from typing import List
-from .test_utils import assert_yaml_equal_except, call_workflow_chat_service, make_service_input, print_response_details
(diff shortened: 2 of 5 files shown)


hanna-paasivirta:
> Well done @hanna-paasivirta !
> 
> So was the problem that step names were getting sanitised beforehand, but not in the edge source/target references?

Sort of, that's the fix at least – the edges are now sanitised. But the actual problem is that the model loses track of quotation marks in its answer. This could be improved by new models, or if we can come up with another model output format strategy where it needs to remember fewer things.

PR #288, opened 2025-09-09
Job_chat: Add answer streaming

### LINKED ISSUE #274: Support streaming interface to Assistant chat
- Better stabilty (fewer timeouts)
- Better UX (more information for users so long waits are more forgivable)

We need to expose eventing (ie, feedback what is happening before tokens start streaming back) and streaming tokens from the Apollo end, then plug Lightning into it.

This is becoming increasingly important as the assistant is slowing down and timing out more and more.

hanna-paasivirta:
## Short Description

Adds streaming/status support to job_chat and workflow_chat. These can now emit custom events (in addition to logs) that are streamed to the frontend via WebSocket. After some work on the Lightning side, this will allow users to see progress instead of waiting for the complete response and enable long running operations without timeout.

Fixes #274 

## Implementation Details

Status updates are sent through apollo/services/util.py 

`job_chat` & `workflow_chat `can now:
- Stream text answers in chunks; send status updates; send the final payload as before once finished (containing the response, history, API usage etc.)
- If using suggested code in job_chat, the service will stream the explanatory answer, but send the suggested code only in the final payload
- Backwards compatible: set `stream = false`

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,7 @@
 {
   "name": "apollo",
   "module": "platform/index.ts",
-<<<<<<< HEAD
-  "version": "0.15.0",
-=======
   "version": "0.16.0",
->>>>>>> 9d7a704 (version: 0.16.0)
   "type": "module",
   "scripts": {
     "start": "NODE_ENV=production bun platform/src/index.ts",


services/job_chat/job_chat.py
```diff
@@ -20,6 +20,21 @@
 
 logger = create_logger("job_chat")
 
+def send_status(message: str):
```
josephjclark, line 23: This stuff will need to be shared by multiple services - so we need to abstract this out into a higher level util function. We might need a helper which accepts the logger (or we just pass it in every time)

josephjclark:
TODO:
- Restructure the emitted event objects to be as close to Anthropic's API as possible

platform/src/middleware/services.ts
```diff
@@ -47,6 +49,75 @@ export default async (app: Elysia, port: number) => {
...
+                isClosed = true;
+              }
+            };
+
+            const onLog = (log: string) => {
+              sendSSE("log", log);
+            };
+
+            const onEvent = (type: string, payload: any) => {
+              // Send events in Anthropic SSE format
```
josephjclark: Probably a claude artifact (and so my fault), but this is an unhelpful comment in the wrong place

josephjclark:
@hanna-paasivirta just thinking aloud, but what about a helper API like this:
```
start_stream()
send_thinking(thinking_note) # for think events, creates one index, sends setup events if it's the first
send_text(chunk) # text events, creates another index
end_stream() # (closes all active content blocks)
```

These functions would need to be stateful and need to track which content blocks are in flight. But I don't think that stuff is super difficult? You could unit test by intercepting the stdout events just to make sure it all works.

We'd need to put a day aside to do the work. It's the sort of thing that if we don't do now, we never will. And I'd certainly like to get this more readable before we launch

PR #306, opened 2025-11-13
Add adaptor function docs upload and search service

### LINKED ISSUE #285: Adaptor documentation summarisation
The issue:
The adaptor documentation is too long: the longest we have come across so far is 900k characters (maybe the length of a couple of PhD theses). 

Why this is a problem:
At worst, this doesn't fit in the prompt and we have to truncate it. We have also noticed in testing that this impairs the model's ability to follow detailed instructions (although we didn't conduct an experiment to check what max length would be safe).

What we've done so far:
We tried to summarise them in a programmatic way (extracting sections, public functions etc) but this did not seem to work reliably as the docs are in different formats.

Todo: 
First:
- One more check for whether there is a simpler solution for this. Check with Joe if some of the longer docs look like what we expect, and if they could be omitted given we already have some adaptor documentation in the prompt.
     - e.g. http 3.1.11: LLMs claim there is no relevant info about the adaptor in the 900k characters, only generic imports that the LLM would already be pretty familiar with.

If we decide we need an LLM-based summarisation service: 
- Set up a service that uses a simple LLM call to summarise text
- It needs to be able to split text that is longer than the context limit into batches, or potentially use a model with a very long context limit (e.g. Gemini). Consider cost as well, although summarisation may require some reasoning in many cases.
-  Only generate once and use a database to store the summaries. This means we need to address issue #284  first.
- Work with Elias to see when we could trigger the summarisation process to minimise user waiting time. E.g. immediately when the user clicks on "AI Assistant" for a workflow that has an adaptor that is not in the database yet.
- Summarise top adaptors before shipping.
- Make the service generic: we might need to summarise all sorts of things in the future, e.g. other documentation, user uploaded docs.

hanna-paasivirta:
## Short Description

Adds `upload_adaptor_docs` and `search_adaptor_docs` services to save version-specific adaptor documentation to Postgres. This will solve the current problems with adding the full TypeScript definitions to the prompt directly. It reduces the adaptor documentation prompt section from at worst 900k characters to just a list of signatures (max possibly 6k characters).

Fixes #285 

### upload_adaptor_docs
Fetches and stores adaptor function docs in PostgreSQL:
- Calls `adaptor_apis` service to fetch JSDoc-generated documentation
- Filters to public functions and external/common functions from language-common
- Stores function signatures, parameters, examples, and descriptions in PostgreSQL
- Namespaces by adaptor name and version
- Returns function count and function list if available. For FHIR-4, which has many functions, this could take 4.5-10.5s.
- If the documentation is not available, we get a response like this:
`{"docs":{},"errors":["@openfn/language-http@3.1.11"]` 
This took 247ms.

### search_adaptor_docs
Queries adaptor function documentation from PostgreSQL:
- Supports multiple query types: function list, signatures, specific function, or all functions
- Version resolution supports "latest" to auto-fetch most recent upload
- Natural language output format for LLM consumption
- Full-text search on function signatures

### job_chat integration
Improved system prompt with adaptor function signatures:
- Replaces previous TypeScript definition approach with database-backed retrieval
- Fetches function signatures from PostgreSQL based on user's adaptor
- The list of function signatures is added to the prompt

Example of the new adaptor functions string in the job_chat prompt:
`'The user is using the OpenFn @openfn/language-http@7.2.5 adaptor. Use functions provided by its API.\n\nThese are the available functions in the adaptor:\n\n\narrayToString(arr, separator)\nas(key, operation)\nchunk(array, chunkSize)\ncombine(operations)\ncursor(value, options)\ndataPath(path)\ndataValue(path)\n\ndel(path, options)\neach(dataSource, operation)\nfield(key, value)\nfields(fields)\nfn(func)\nfnIf(condition, operation)\nget(path, options)\ngroup(arrayOfObjects, keyPath, callback)\nhumanProper(str)\nlastReferenceValue(path)\nmap(path, callback)\nmerge(dataSource, fields)\nparseCsv(csvData, parsingOptions, callback)\nparseXML(data, script)\npatch(path, data, options)\npost(path, data, options)\nput(path, data, options-)\nrequest(method, path, options)\nscrubEmojis(text, replacementChars)\nsourceValue(path)\nsplitKeys(obj, keys)\ntoArray(arg)\ndecode(base64Data)\nencode(data)\nuuid()\n'`

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.changeset/light-moose-ring.md
+++ b/.changeset/light-moose-ring.md
@@ -1,0 +1,8 @@
+---
+"apollo": minor
+---
+
+- Add `load_adaptor_docs` and `search_adaptor_docs` services, which use postgres
+  to cache and search docs data
+- Use new adaptor docs lookup in `job_chat`, for faster, leaner and better docs
+  lookup


josephjclark:
1. On upload, return a sensible status code (eg, 200, 201 or 204)
2. On upload, it's cool to fail quickly for tags that don't exist (maybe we'll optmise the npm package)
3. In job chat prompt assembly, if search returns nil,  then a) raise a sentry warning and b) don't attach any adaptor docs to the prompt
4. Over in lightningville, lightning will call the uploader when a chat session starts (or something) to prime the DB before the user asks a question. We'll need to raise an issue for this

services/job_chat/prompt.py
```diff
@@ -278,23 +281,53 @@ def generate_system_message(context_dict, search_results):
...
+            f"<adaptor>The user is using the OpenFn {context.adaptor} adaptor. Use functions provided by its API.\n\n"
         )
 
         try:
-            adaptor_docs = describe_package(context.adaptor)
-            for doc in adaptor_docs:
-                adaptor_string += f"Typescript definitions for doc {doc}"
-                adaptor_string += adaptor_docs[doc]["description"]
+            # Get database connection
+            db_url = os.environ.get("DATABASE_URL")
```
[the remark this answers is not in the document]
josephjclark: And actually I note this logic is duplicated across prompt and old_prompt - can we abstract it out?

Because of the react rewrite I fear we'll have to maintain the old prompt for a while longer

services/load_adaptor_docs/README.md
```diff
@@ -0,0 +1,253 @@
...
+  - Full form: `"@openfn/language-kobotoolbox"`, `"@openfn/language-dhis2"`, etc.
+- **`version`** (string): Adaptor version (e.g., `"4.2.7"`)
+
+**Optional:**
+- **`skip_if_exists`** (boolean): If `true`, checks database first and returns existing data if found. If `false` (default), always fetches and replaces docs.
+- **`POSTGRES_URL`** (string): PostgreSQL connection string (falls back to environment variable)
+
+### Behavior modes:
+
+**Default mode (`skip_if_exists: false` or not provided):**
```
josephjclark: My instinct is that this should be `true` by default - I think it's quite rare we want to force a reload.

services/util.py
```diff
@@ -96,39 +98,69 @@ def apollo(name, payload):
...
+    if db_url:
+        return psycopg2.connect(db_url)
+    return None
 
-    Accepts:
-    - "@openfn/language-http@3.1.11" -> ("@openfn/language-http", "3.1.11", "@openfn/language-http@3.1.11")
-    - "http@3.1.11" -> ("@openfn/language-http", "3.1.11", "@openfn/language-http@3.1.11")
 
-    Raises ApolloError if version is not provided.
+class AdaptorSpecifier:
```
josephjclark: That's lovely - I think that will make handling adaptor strings a lot cleaner :ok_hand:

josephjclark (commented):
Still a couple of things bothering me - see the latest comments.

Running locally I'm very comfortable with the performance. It think it's about par for new adaptor versions with main - around 2-3 seconds to download (and process) the docs. I will test again on staging but I think we're onto a winner without needing to add any cache warming stuff.

josephjclark (approved):

PR #357, opened 2026-01-20
add branching explanation to prompt

### LINKED ISSUE #352: Workflow chat:  model doesn't understand edges
This generated workflow is invalid:

Image

The Consolidate All Data step thinks it'll run when both its parent steps have completed.

While this is a feature we'd like some day, it's not how it works! Every edge is executed simultaneously with no waiting.

We should update the prompt to be a little more clear/strict on this

hanna-paasivirta:
## Short Description

Added clarification to the workflow_chat prompt explaining that edges execute independently and simultaneously. When multiple edges target the same job, that job runs once per incoming edge rather than waiting for all parent jobs to complete and merging their state. This fixes incorrect model explanations of workflows where parallel jobs converge to a single downstream step.

I tested with Taylor's input in the issue and this seems to work. 

Fixes #352

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.changeset/yummy-ties-tap.md
+++ b/.changeset/yummy-ties-tap.md
@@ -1,0 +1,5 @@
+---
+"apollo": patch
+---
+
+workflow_chat: improve the model's understanding of edge branching behaviour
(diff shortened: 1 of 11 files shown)


josephjclark (approved):

josephjclark:
Great. Can we add a changeset please?

PR #407, opened 2026-03-09
feat: reverse streaming order so changes arrive before text

elias-ba:
## Short Description

Reverses the streaming order in both job_chat and workflow_chat so that structured changes (code edits / workflow YAML) are sent to the client **before** the text explanation streams, enabling a better UX where users see results first.

## Implementation Details

When the AI assistant streams a response, the text explanation ("I've updated your workflow to do X, Y, Z") appeared first, and then seconds later the actual changes landed on the canvas or code editor. Users would read about changes that hadn't happened yet.

The fix is to reverse the JSON field order via the assistant prefill so that structured data (code_edits or YAML) is generated first and text second. During streaming, the structured data is buffered silently. Once the delimiter between fields is detected, the structured data is parsed and sent to the client as a new custom SSE event called `changes`, and then the text explanation streams normally.

For job_chat, the raw code_edits are patches (replace/rewrite), so a new `_resolve_code_edits()` method applies them to the user's current code at stream time to produce the final code the client needs for diff preview. This is a best-effort version of `apply_code_edits()` without error correction to avoid blocking the stream.

Both services also add `_unescape_json_string()` to convert JSON escape sequences (\\n, \\") back to real characters during streaming, since Claude generates text inside a JSON string value.

This works together with a companion Lightning PR that handles the `changes` event on the client side.

## AI Usage

- [ ] Strategy / design


hanna-paasivirta (approved):
Looks great, thank you Elias. My tests look good, including the correction loop to job code edits.

PR #426, opened 2026-03-25
fix: skip sending null YAML as streaming changes event

### LINKED ISSUE #427: Streaming sends "null" as YAML changes event for informational responses
When the AI assistant responds to a simple greeting or informational question (no workflow changes), the streaming code sends a `changes` event with `{yaml: "null"}`. This causes the frontend to display "null" as generated workflow code with an Apply button.

**Root cause:** The model returns `{"yaml": null, "text": "..."}`. During streaming, the raw text `null` is unescaped via `_unescape_json_string` into the string `"null"`, which is truthy and passes the `if yaml_value:` check.

**Expected behavior:** No `changes` event should be sent when there are no workflow YAML changes.

elias-ba:
## Short Description

Skip sending "null" as a YAML streaming changes event when the model returns no workflow changes.

Fixes #427

## Implementation Details

When the model returns `{"yaml": null, "text": "..."}` for informational responses (e.g. simple greetings), the streaming code unescapes JSON `null` into the string `"null"` via `_unescape_json_string`. This truthy string passes the `if yaml_value:` check and gets sent as a `changes` event, causing the frontend to display "null" as generated workflow code with an Apply button.

The fix adds a check that `yaml_value != "null"` before sending the `changes` event. This specifically handles the semantic mismatch where JSON `null` becomes the truthy string `"null"` through the unescaping step.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.changeset/empty-oranges-turn.md
+++ b/.changeset/empty-oranges-turn.md
@@ -1,0 +1,5 @@
+---
+"apollo": patch
+---
+
+fix empty yaml key handling after streaming order reversal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # apollo
 
+## 0.21.1
+
+### Patch Changes
+
+- workflow_chat: Fix streaming sending "null" as YAML changes event for
+  informational responses (#427)
+
 ## 0.21.0
 
 ### Minor Changes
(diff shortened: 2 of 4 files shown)


hanna-paasivirta:
I fixed some misleading prompts about the required format for the empty YAML key, and accordingly the streaming handling in read-only mode so it’s not stuck waiting for the text start delimiter which won’t be generated.

hanna-paasivirta (approved):
Looks great, thank you Elias. I added a fix for some misleading prompts, and accordingly the streaming handling in read-only model as well.

PR #429, opened 2026-03-26
feat: router outputs target job_key when routing to job_code_agent

### LINKED ISSUE #428: Router should output target job_key when routing to job_code_agent
## Problem

When the global chat routes a request to `job_code_agent`, it relies on the `page` URL to identify which job to extract context from and stitch code back into. When the user is on the workflow overview (no step name in the page URL), `_route_to_job_chat` can't identify the target job — `matched_job_key` stays `None`, no job context is extracted, and code stitching doesn't happen.

This means job code requests from the workflow overview silently fail — the model says "I've updated the code" but the YAML is unchanged.

## Root Cause

The router LLM already sees the full workflow YAML and the user's message to make the routing decision, but it only outputs `destination` and `confidence`. When routing to `job_code_agent`, it doesn't specify *which* job.

## Proposed Fix

1. Add optional `job_key` field to `RouterDecision`
2. Update the routing system prompt to also return `"job_key": "Fetch-data"` when routing to `job_code_agent`
3. In `_route_to_job_chat`: use `step_name` from page first, fall back to `decision.job_key` when page has no step name

This is zero additional LLM cost — the router is already making an API call, we just ask it to output one more field.

elias-ba:
## Short Description

When routing to job_code_agent, the router now also outputs the target job_key so that job code changes work even when the user is on the workflow overview (no step name in the page URL).

Fixes #428

## Implementation Details

When the user is on the workflow overview and asks the global assistant to modify job code, the router correctly identifies the request as a job_code_agent task. However, `_route_to_job_chat` relies on the `page` URL to identify which job to extract context from and stitch code back into. Without a step name in the URL, `matched_job_key` stays `None` and code stitching silently fails.

The router LLM already sees the full YAML and user message to make the routing decision. This PR extends it to also output the target `job_key` (matching a key from the workflow YAML) when routing to `job_code_agent`.

Changes:
- Add optional `job_key` field to `RouterDecision`
- Update routing prompt to require `job_key` when destination is `job_code_agent`
- Parse `job_key` from LLM response
- `_route_to_job_chat` falls back to `router_job_key` when `get_step_name_from_page` returns nothing

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/services/global_chat/PAYLOAD_SPEC.md
+++ b/services/global_chat/PAYLOAD_SPEC.md
@@ -118,7 +118,7 @@ This document defines the input and output payload structure for the Global Agen
 
 - **`response`** (string): The main text response from the agent.
 
-- **`attachments`** (array): Artifacts produced during this turn. Each entry has a `type` and `content` field. An empty list `[]` means no artifacts were produced (e.g. a purely informational response). Currently supported types: `workflow_yaml`.
+- **`attachments`** (array): Artifacts produced during this turn. Each entry has a `type` and `content` field. An empty list `[]` means no artifacts were produced (e.g. a purely informational response). Currently supported types: `workflow_yaml`, `job_code`. When both are present, `job_code` contains the suggested code for a specific job and `workflow_yaml` contains the full YAML with the code stitched in.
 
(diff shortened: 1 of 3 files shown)


hanna-paasivirta (approved):
I think getting the router to predict the job key is a really good idea. I'll monitor this to understand if we need additional logic or safeguards around the prediction.

PR #431, opened 2026-03-26
feat: return job_code attachment when routing to job_code_agent

### LINKED ISSUE #430: Router should return job_code attachment when routing to job_code_agent
## Problem

When the global chat router delegates to `job_code_agent`, the response only contains a `workflow_yaml` attachment (full YAML with code stitched in). The frontend renders this as a "Generated Workflow" YAML card, even when the user is viewing a specific job in the code editor.

The normal (non-global) `job_chat` returns `suggested_code` directly, which the frontend renders as a code diff in Monaco. The global chat loses this granularity by wrapping everything into `workflow_yaml`.

## Proposed Fix

In `_route_to_job_chat`, return both attachment types when a job code change was made:
- `{type: "job_code", content: suggested_code}` — for rendering as a code diff
- `{type: "workflow_yaml", content: updated_yaml}` — for applying to the canvas

Then on the Lightning side, `build_global_message` picks `job_code` over `workflow_yaml` for the message's `code` field, so the frontend renders a code diff instead of a YAML card.

elias-ba:
## Short Description

When the router delegates to job_code_agent, include the suggested code as a separate job_code attachment alongside the workflow_yaml attachment.

Fixes #430

## Implementation Details

When the global chat routes to job_code_agent, the response currently only contains a workflow_yaml attachment (full YAML with code stitched in). The frontend renders this as a "Generated Workflow" YAML card, even when the user is viewing a specific job in the code editor and would expect to see a code diff.

The normal (non-global) job_chat returns suggested_code directly, which the frontend renders as a code diff in Monaco. The global chat was losing this granularity by wrapping everything into workflow_yaml.

Now _route_to_job_chat returns both attachments when suggested_code is available:
- {type: "job_code", content: suggested_code} — for rendering as a code diff
- {type: "workflow_yaml", content: updated_yaml} — for applying to the canvas

The Lightning side can then pick job_code over workflow_yaml for the message code field, so the frontend renders a code diff instead of a YAML card when appropriate.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/services/global_chat/PAYLOAD_SPEC.md
+++ b/services/global_chat/PAYLOAD_SPEC.md
@@ -118,7 +118,7 @@ This document defines the input and output payload structure for the Global Agen
 
 - **`response`** (string): The main text response from the agent.
 
-- **`attachments`** (array): Artifacts produced during this turn. Each entry has a `type` and `content` field. An empty list `[]` means no artifacts were produced (e.g. a purely informational response). Currently supported types: `workflow_yaml`.
+- **`attachments`** (array): Artifacts produced during this turn. Each entry has a `type` and `content` field. An empty list `[]` means no artifacts were produced (e.g. a purely informational response). Currently supported types: `workflow_yaml`, `job_code`. When both are present, `job_code` contains the suggested code for a specific job and `workflow_yaml` contains the full YAML with the code stitched in.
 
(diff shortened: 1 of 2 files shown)


hanna-paasivirta (approved):
I think this is good for the current purpose of making the experimental prototype much more functional. This won't solve the front-end diff problem in all scenarios (e.g. planner called job code agent; planner called job code agent several times but the user is viewing a particular step) and this is ok at this stage.

PR #466, opened 2026-04-20
Streaming bugs

### LINKED ISSUE #461: Order of streamed artefacts doesn't seem right
When streaming, workflow_chat and job_chat may not render changes before text (even though we want to render changes before text)

This might be intermittent, but seems to be happening regularly

hanna-paasivirta:
## Short Description

Fixes streaming bugs introduced with the switch to structured outputs that led to job code/YAML not being streamed before the text answers.

Fixes #461 

### `EVENT:changes` wasn't firing in `workflow_chat`

When we switched to structured outputs and dropped answer prefills, the buffered content became a fragment of an unclosed object like `{"yaml": "...",` — which always fails to parse, so `send_changes` was silently skipped and yaml/code only arrived in the final `complete` payload. Now, we strip the trailing comma + whitespace, append `}` to make it valid JSON, and extract the field by key. `job_chat` does this already (but this PR adjusts it slightly to clean up dead code and match the approach in the two services).

### The field-boundary delimiter was too literal

The code matched the exact string `'"text": "'` / `'"text_answer": "'` with a hardcoded space. Anthropic's structured outputs emits JSON with inconsistent whitespace . `workflow_chat` happened to come back pretty-printed, but `job_chat` comes back compact (`"text_answer":"`). `job_chat`'s delimiter never matched, so text never streamed at all and everything appeared at once. We've now replaced the literal match with regex `r'"text"\s*:\s*"'` so whitespace doesn't matter.

## AI Usage

- [x] Strategy / design
- [x] Other
- [ ] I have not used AI


josephjclark (approved):
Works great, thanks @hanna-paasivirta

PR #478, opened 2026-04-27
Service test planning

hanna-paasivirta:
## Short Description

Updated section 2) service tests in agent-team-architecture-plan/2-service-tests.md 

## AI Usage

- [x] Strategy / design
- [x] Other
- [ ] I have not used AI

### DIFF
--- a/agent-team-architecture-plan/2-service-tests.md
+++ b/agent-team-architecture-plan/2-service-tests.md
@@ -185,10 +185,7 @@ services/testing/
   yaml_assertions.py    # YAML structural helpers (unit-tier safe; owned by unit tier)
 ```
 
-`fixtures.py` holds:
-
-- `make_global_chat_payload`, `make_workflow_chat_payload`, `make_job_chat_payload`
-- Pytest fixtures: `mock_anthropic`, `test_hooks_factory`, `fake_api_key`
+`fixtures.py` registers pytest fixtures via `pytest_plugins = ["testing.fixtures"]` in the root conftest. Today: just `test_hooks_factory` (builds a `test_hooks` dict, threading in a `MockAnthropic`'s httpx client). Add more fixtures here as service tests need them.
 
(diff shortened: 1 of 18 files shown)


agent-team-architecture-plan/2-service-tests.md
```diff
@@ -85,71 +85,110 @@ def build_anthropic_client(api_key: str, test_hooks: Optional[dict] = None) -> A
...
-        mc = MockAnthropicClient.script([resp1, resp2, resp3])   # multi-turn
-        mc = MockAnthropicClient.streaming(events=[...])         # SSE
-
-    After the call:
-        mc.requests            # list[RecordedRequest]
-        mc.last_request.json["messages"]
-        mc.last_request.headers["x-api-key"]
+        mock = MockAnthropic()
+        mock.set_response(r"haiku", "sure, here's a haiku")
+        mock.set_response(r"create workflow", tool_use("call_workflow_agent", {...}))
```
josephjclark: Not sure I understand this bit

josephjclark: Oh, is it changing the response structure to be a like a strucuted tool output, rather than free text?

So the regular response is like `{ content: "here's a haiku" }`, but the tool response is like `{ type: "tool_use", content: { rag_results: [] } }` or whatever

It might be nicer to have functions like `mock.set_tool_response(type, input, output)` , which give us more control when simulating these particular requests.

But we can resolve that in code, maybe we don't need to interfere with the spec yet

hanna-paasivirta: By set_tool_response(type, input, output), do you mean one call that handles the whole round trip like "when the model decides to call tool x with input y, the tool returns z"? Now they're separate as set_response + tool_use() for the model request and tool_stubs for what the tool returns

josephjclark: Yeah one call for the round trip is what I was thinking of

This approach with the `tool_use` call just feels weird to me.

But maybe we should just generate with this approach and take a close look at it, then maybe experiment with other approaches. Sometimes it's easier to make these decisions against a diff.

PR #479, opened 2026-04-27
Testing architecture unit tests

hanna-paasivirta:
## Short Description

Implements the unit test architecture according to the testing plan section 1 apollo/agent-team-architecture-plan/1-unit-tests.md

## AI Usage

- [x] Strategy / design
- [x] Other
- [ ] I have not used AI


services/testing/README.md
```diff
@@ -0,0 +1,22 @@
...
+This directory is on the Python path via `pyproject.toml`
+(`pythonpath = ["services"]`), so tests import as
+`from testing.yaml_assertions import assert_yaml_equal_except`.
+
+## Modules
+
+- `yaml_assertions.py` — pure-function YAML structural assertions, safe for
+  every tier (unit included).
+
+## Why under `services/` and not a top-level `tests/`?
```
josephjclark, line 14: hmm. Aren't we better off fixing the imports or even restructuring?

Looking over the codebase there's actually quite a lot in `services` which isn't really a service. But I like that clean divide where services folders are expected to have a strong structure convention

I wonder if we should do a bigger restructure like:
```
/py
  - /services
  - /util
    - /entry.py
    - /models.py
```

Then you'd configure imports to run from py. And shouldn't we be able to do relative imports within a service?

This feels like quite a bit of work but we're adding so much non-service python code  that I'm keen to get it right.

Let's discuss next week. Might be good to pull Elias in too

services/workflow_chat/tests/unit/test_isolation_guard.py
```diff
@@ -0,0 +1,17 @@
+"""Smoke tests for the unit-tier isolation guard in the root conftest.py."""
```
josephjclark: I don't understand what this is - do you?

services/workflow_chat/tests/unit/test_workflow_chat_client_extract.py
```diff
@@ -0,0 +1,28 @@
+def test_extract_job_codes_preserves_real_code(workflow_chat_client):
```
josephjclark: I don't understand this either. Who is calling ths function?

If this is tests against `extract_and_preserve_components`, cool! But this file should be importing and creating a client itself

services/workflow_chat/tests/unit/gen_project/test_prompt_build.py
```diff
@@ -0,0 +1,49 @@
+from workflow_chat.gen_project_prompt import build_prompt
```
josephjclark, line 1: This file feels good: it's importing a single function and setting up a bunch of independent unit tests on it.

The file name is pretty awkward though. Should we introduce more folders here? `workflow_chat/tests/unit/gen_project/test_prompt_build.py`

services/workflow_chat/tests/unit/test_workflow_chat_client_sanitize.py
```diff
@@ -0,0 +1,40 @@
+def test_sanitize_job_names_removes_diacritics(workflow_chat_client):
+    yaml_data = {
+        "jobs": {
+            "job1": {"name": "Café München"},
+            "job2": {"name": "Naïve résumé"},
+        }
+    }
+
+    workflow_chat_client.sanitize_job_names(yaml_data)
```
josephjclark: Again it's wierd to me that has a client passed into it, and I can't find a caller in the codebase

services/global_chat/tests/test_utils.py
```diff
@@ -3,10 +3,20 @@
 import yaml
 import tempfile
 import subprocess
-import difflib
 from pathlib import Path
 from typing import Dict, List, Optional, Any
 
+# YAML helpers live in `testing.yaml_assertions`; re-exported so existing
```
josephjclark: I find this comment a bit sus. Why do we need to explain a basic import like this?

josephjclark:
Tests are now running against this branch, which is cool. I'm actually a bit surprised by that because I thought the action had to be on main for hooks to trigger. I must be getting  my wires crossed there.

Anyway it's failing on the poetry install step. I don't have time to investigate this afternoon - I'll sort it next week.

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,34 @@
+name: Unit tests
+
+# Runs when a pull request is created or its head is updated
+# https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
+on: pull_request
+
+jobs:
+  unit:
+    name: Python unit tests
```
josephjclark, line 9: And of course this only has python unit tests, not the typescript ones. Another job for next week

services/workflow_chat/tests/unit/test_workflow_chat_client_sanitize.py
hanna-paasivirta: Are you saying we should change the service code so the function isn't tied to the class? 

And what do you mean you can't find a caller?

services/workflow_chat/tests/unit/client/test_extract.py
```diff
@@ -0,0 +1,39 @@
+from workflow_chat.workflow_chat import AnthropicClient
+
+
+def _make_client():
```
josephjclark: This is a strange pattern

I get it, but it's not pretty

What's happening is we're using a weird python syntax to create a client instance without triggering the constructor. We have to do this because the constructor creates an Anthropic client and blows up in test.

But it's all backwards that we're working around this. We're in total control of the code.

The best option I think is to make `extract_and_preserve_components` a static method. Now you don't need to instantiate a client instance to use it.

We could also modify the constructor to disable client creation, or even better accept a mock client instance, just for tests. We may need to do this one day.

Finally we could declare `extract_and_preserve_components` as a standard python function (not in the class). Since it doesn't use any class state it might make more sense. Then we either remove it from the class, or just proxy into it if `self.extract_and_preserve_components` is a useful thing to have.

services/workflow_chat/tests/unit/test_workflow_chat_client_sanitize.py
josephjclark: > Are you saying we should change the service code so the function isn't tied to the class?

I've lost the diff, but I don't think  I was saying that. But also that's a good solution!

> And what do you mean you can't find a caller?

Just that no-one in the code seemed to be calling the function

services/workflow_chat/tests/unit/client/test_extract.py
hanna-paasivirta: I have split it into a client folder, and made the extraction and sanitization into static methods. 

We don't really have simple unit tests yet in job_chat or global_chat. Should I add a few under this new architecture?

josephjclark:
Replaced with https://github.com/OpenFn/apollo/pull/486

PR #481, opened 2026-04-29
Fix bun test payload echo failure

### LINKED ISSUE #463: `bun test` fails
I would like `bun test` to be passing again :(

alceops:
Fixes #463

## Summary
- Keeps the service payload echo behavior aligned with the server contract: JSON POST bodies still receive the internal `session_id` before invoking service handlers.
- Tightens the echo regression so it explicitly expects the session id instead of stripping it out.
- Preserves empty-body safety with `ctx.body ?? {}`.

## Verification
- `bun test --timeout 20000 platform/test/server.test.ts -t 'echoes back an object with a session id'` → pass
- `bun test --timeout 20000 platform/test/server.test.ts` → 7 passed
- `git diff --check`

Notes on local setup: this runner used Bun plus a local Poetry/Python shim with the minimal dependencies needed by the existing tests. No secrets, service API keys, deploys, account/payment setup, or production data were used.

### DIFF
--- a/platform/src/middleware/services.ts
+++ b/platform/src/middleware/services.ts
@@ -34,8 +34,7 @@ export default async (app: Elysia, port: number) => {
       // simple post
       app.post(name, async (ctx) => {
         console.log(`POST /services/${name}: ${ctx.uuid}`);
-        const payload = ctx.body;
-        payload.session_id = ctx.uuid;
+        const payload = ctx.body ?? {};
         const result = await callService(m, port, payload as any);
 
         if (isApolloError(result)) {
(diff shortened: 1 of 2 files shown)


platform/src/middleware/services.ts
```diff
@@ -34,8 +34,7 @@ export default async (app: Elysia, port: number) => {
       // simple post
       app.post(name, async (ctx) => {
         console.log(`POST /services/${name}: ${ctx.uuid}`);
-        const payload = ctx.body;
-        payload.session_id = ctx.uuid;
```
josephjclark, line 38: We need this session id! Or at least we will soon.

Please update the test to expect it, rather than stripping it out.

Thank you!

josephjclark (changes requested):

alceops:
Thanks — updated to keep `session_id` in the service payload and changed the echo regression to expect it instead of stripping it.

Verification (local shim: Bun + minimal Poetry/Python deps):
- `bun test --timeout 20000 platform/test/server.test.ts -t 'echoes back an object with a session id'` ✅
- `bun test --timeout 20000 platform/test/server.test.ts` ✅ 7 passed
- `git diff --check` ✅

josephjclark (approved):
Thank you @alceops !!

PR #486, opened 2026-05-07
Merge unit tests

josephjclark:
This PR rebases #479 against latest main and add a couple of tweaks ready for merging, including:
- dropping the spec files, which we don't want on main
- fixing the github action


josephjclark:
Ok, the action is working and the work is rebased. But tests are failing - mostly I think because an api key is needed on tne anthropic client. This shouldn't really be needed so we'll investigate solutions a bit later.

josephjclark:
Oh nice, the tests are working! I've just had to be a bit more selective about the tests we run

PR #487, opened 2026-05-07
Return error if the chat answer is empty

### LINKED ISSUE #484: Chat services return success with empty response when LLM output is unparseable or empty
## What's broken

Both `workflow_chat` and `global_chat` can return HTTP 200 with `response: ""` (empty string) when the LLM produces output that the service can't use. There is no legitimate "empty response" path, so this silently degrades the contract: callers see a success and have no way to distinguish "the model said nothing useful" from "the model's normal answer".

In `services/workflow_chat/workflow_chat.py`, `split_format_yaml` initialises `output_text = ""` and only fills it from `response_data.get("text", "")` after JSON parsing succeeds. If the LLM returns text that isn't valid JSON (or JSON without a `text` field), the `except` branch logs an error and `output_text` stays empty. The wrapper still returns `{"response": "", ...}` with status 200.

In `services/global_chat/planner.py`, `final_text = ""` initialises and is filled by `_extract_text(response)` only on `end_turn`, on max_tool_calls exit, or on an unexpected `stop_reason`. If the model produces a response with no text blocks (only `tool_use` or `thinking`) and the loop exits without `end_turn`, `_extract_text` returns empty. The service logs `Loop exited without end_turn` but still returns 200 with `response: ""`.

## How it surfaced

In Lightning, the AI Assistant background worker treats Apollo's 200 as success and tries to persist the assistant turn. The `ChatMessage` changeset requires `content` (1 to 10,000 characters), so the insert fails with `content: "can't be blank"`. The user-side message stays stuck in `:processing` and the user sees no response and no error indication.

Surfaced as Sentry alert LIGHTNING-1MP and tracked downstream at OpenFn/lightning#4710.

## What to fix

When the LLM output can't be parsed (workflow_chat) or produces no text content (global_chat), the service should raise `ApolloError` (or include an explicit error signal in the response body), not return 200 with empty text.

Both services already use `ApolloError` for other failure modes (auth, rate limit, connection), and Lightning's `handle_error_response` already routes those into the error tuple cleanly, so the fix should fit the existing shape without new contracts.

hanna-paasivirta:
## Short Description

`workflow_chat`, `global_chat`, and `job_chat` could return HTTP 200 with response: `""` when the LLM produced no usable text. Lightning treated the 200 as success leaving the user-side message stuck in `:processing`. Each service now raises ApolloError when there's no usable text, so Lightning's handle_error_response routes it into the error tuple.

Fixes #484 

### ApolloError usage
Each service raises `ApolloError(502, ...)` when text output is empty after the existing retry/loop logic. Type is `OUTPUT_TRUNCATED` for `stop_reason=max_tokens`, `EMPTY_OUTPUT` otherwise. Code and types are new values within the existing `ApolloError` shape.

### Sentry tracking
We attach Sentry reporting before each raise. `entry.py`'s existing capture_exception records one event per failed request.

### Tests
Added temporary tests in separate test files that won't run automatically. This is to avoid conflicts from our testing architecture which is currently in progress. These tests will need to be rewritten to fit into the new tests.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # apollo
 
+## 1.1.1
+
+### Patch Changes
+
+- Improve error reporting when the model returns an empty message
+
 ## 1.1.0
 
 ### Minor Changes
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "apollo",
   "module": "platform/index.ts",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "type": "module",
   "scripts": {
     "start": "NODE_ENV=production bun platform/src/index.ts",
--- a/services/global_chat/planner.py
+++ b/services/global_chat/planner.py
@@ -7,6 +7,7 @@ from typing import List, Dict, Optional
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass
 from anthropic import Anthropic
+import sentry_sdk
 
 import sys
 from pathlib import Path
@@ -208,6 +209,8 @@ class PlannerAgent:
                         logger.warning(f"Unexpected stop_reason: {response.stop_reason}")
                         break
 
+                except ApolloError:
+                    raise
                 except Exception as e:
                     logger.exception("Error in tool-calling loop")
                     raise ApolloError(500, f"Tool execution error: {str(e)}")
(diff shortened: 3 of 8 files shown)


josephjclark (approved):
I've not been able to repro but this looks good.

Interesting use of mocking. Let's consider this pattern more when we come to the proper integration tests

I've added a changeset and will prep the release

PR #515, opened 2026-06-08
Workflow_chat: Move placeholder string replacement to before YAML stream event

### LINKED ISSUE #4828: AI assistant emits placeholder ids/code (`__ID_JOB_…__`, `__CODE_BLOCK_…__`) instead of UUIDs/real code in generated workflows
## Summary

The AI assistant (workflow template generation, via the Apollo service) sometimes produces a workflow whose nodes carry **unsubstituted template placeholders** instead of real values — symbolic ids like `__ID_JOB_<name>__` / `__ID_EDGE_<name>__` and code bodies like `__CODE_BLOCK_<name>__`. When such a workflow is applied to the collaborative Y.Doc and saved, the placeholder ids reach the DB as `:binary_id` values and fail.

Until recently this **crashed the collaborative Session** (see #4816). With the changeset validation added there, it no longer crashes — the malformed id now surfaces as a **toast/validation error** to the user. **This issue exists to track the actual root cause: why the AI assistant emits placeholders rather than substituting them, and to confirm when that is fixed.** The crash fix is a safety net, not a fix for the AI behaviour.

## Observed placeholder format

From production crash payloads (the full `save_workflow` payload was captured in the Session crash):

```
jobs:    "id" => "__ID_JOB_Envoyer-dans-DHIS2__", "body" => "__CODE_BLOCK_Envoyer-dans-DHIS2__"
edges:   "id" => "__ID_EDGE_Collecter-les-donnees->Envoyer-dans-DHIS2__",
         "source_job_id" => "__ID_JOB_Collecter-les-donnees__"
```

Crucially the payload is a **mix** of placeholders and real UUIDs — some jobs/edges/triggers already had valid ids (e.g. job `77438a5a-…`, trigger `3ae0637e-…`) alongside the placeholders. This suggests a **partial substitution / partial apply**, not a wholesale failure.

## Where the placeholders come from

The tokens `__ID_JOB_`, `__ID_EDGE_`, `__CODE_BLOCK_` appear **nowhere in the Lightning codebase** (backend or frontend). They originate from the **Apollo AI service** output for `session_type: "workflow_template"` (`Lightning.AiAssistant` → `Lightning.ApolloClient`). Lightning is the consumer: the generated workflow is stored on the chat message `code` and applied to the Y.Doc, then saved via `LightningWeb.WorkflowChannel` → `save_workflow`. Substitution of these placeholders into real UUIDs/code is expected to happen before the workflow is applied/saved; in these cases it didn't (fully).

## Production evidence (Sentry `LIGHTNING-1AF`, `Ecto.ChangeError` slice)

All on `Lightning.Workflows.Job.id`, message `value "__ID_JOB_<name>__" … does not match type :binary_id`:

| When (UTC) | Events | User (internal, `support_user: true`) | Workflow | Release |
|---|---|---|---|---|
| 2026-04-27 12:04–13:47 | 5 | mtuchi dev `4baaf510-2594-4a3a-aada-db4b64bd9a34` | `8a1a736b-b28a-4770-b7e0-8cff31bfcb92` — "WF1 - Aggregate surveillance data from STELab and push to DHIS2" (lock_version 30) | v2.16.2 (1.5.81) |
| 2026-04-08 13:31 | 1 | Taylor Downs `a2744481-743a-46e1-9224-429aef7ea753` | `0e7b96f6-5ac3-41b0-84b3-119bfa515167` — "other name here" (lock_version 3) | v2.16.1 (1.5.79) |

Sample event IDs (2026-04-27): `85fa35f8…`, `a1794e16…`, `5415a14a…`, `ce686a11…`, `d286a2c9…`. (2026-04-08): `3ecfdc24…`.

Notes:
- Both impacted users are **internal OpenFn staff/support accounts**, not external customers.
- The 5 April-27 events are the **same workflow re-saved repeatedly** (consistent with a user retrying a save on a broken AI-generated doc).
- This is bounded to Sentry's 90-day retention window; `LIGHTNING-1AF` goes back to 2026-01-12, so earlier occurrences may have aged out.

## Pointers for investigation

- Reproduce a `workflow_template` AI generation and inspect the raw Apollo response stored on the chat message `code` — does it contain `__ID_*__` / `__CODE_BLOCK_*__` tokens, and under what conditions?
- Determine where substitution is meant to occur (Apollo response assembly vs. Lightning apply path) and why it can be skipped/partial.
- To locate the exact AI chat session for the main case, join on `workflow_id = 8a1a736b-…` + `user_id = 4baaf510-…` around 2026-04-27 12:00–14:00 UTC.

## Related

- #4816 — the crash fix (changeset validation now turns this into a toast instead of a Session crash). This issue tracks the upstream AI behaviour.

hanna-paasivirta:
## Short Description

This PR moves the placeholder string replacement in `workflow_chat` to be applied before we emit the `changes` event containing the generated YAML, so that we never send placeholder strings to Lightning.

Fixes https://github.com/OpenFn/lightning/issues/4828

## Implementation details

The UUID and code placeholders that were sent to Lightning in the `changes` event were generally invisible to the user, because Lightning corrects the interim YAML with the YAML sent in the final payload. However, if the user were to click into a step they might potentially see the placeholders briefly. While this doesn't appear to be causing any problems for users now, it might become problematic with the Global chat, as users might already be viewing the job code when the changes stream in – in which case they might be more likely to see the placeholders. Given this and the errors in Lightning, it makes sense to move the placeholder string replacement steps to occur before we emit the YAML in `changes`.

## AI Usage

- [x] Strategy / design
- [x] Other
- [ ] I have not used AI

### DIFF
--- a/services/workflow_chat/workflow_chat.py
+++ b/services/workflow_chat/workflow_chat.py
@@ -132,6 +132,11 @@ class AnthropicClient:
         if not self.api_key:
             raise ValueError("API key must be provided")
         self.client = Anthropic(api_key=self.api_key)
+        # Holds the finalized YAML sent in the streaming `changes` event. The
+        # final response reuses this exact string rather than finalizing again,
+        # so restore_components runs once and new-component UUIDs stay identical
+        # between the streamed preview and the persisted payload.
+        self._streamed_yaml = None
 
     @staticmethod
     def _unescape_json_string(text):
@@ -205,6 +210,8 @@ class AnthropicClient:
 
             max_retries = 1
             for attempt in range(max_retries + 1):
+                # Reset per attempt so a retry never reuses a prior stream's YAML
+                self._streamed_yaml = None
                 with sentry_sdk.start_span(description="anthropic_api_call"):
                     if stream:
                         logger.info("Making streaming API call")
@@ -231,7 +238,8 @@ class AnthropicClient:
                                     accumulated_response,
                                     text_started,
                                     sent_length,
-                                    stream_manager
+                                    stream_manager,
+                                    preserved_values
                                 )
                         message = stream_obj.get_final_message()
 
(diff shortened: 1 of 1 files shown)


josephjclark:
To test: placeholder ids will always appear in the streamed lightning logs on main. On this PR they go away.

PR #517, opened 2026-06-10
Release: 1.3.0

josephjclark:
- fix placeholders
- update deps


josephjclark:
Suddenly seeing errors in job chat and global chat:
```
The `pinecone-plugin-inference` package has been deprecated. The features from that plugin have been incorporated into the main `pinecone` package with no need for additional plugins. Please remove the `pinecone-plugin-inference` package from your dependencies to ensure you have the most up-to-date version of these features.
```

EDIT: purging the poetry venv resolved it (without a diff). I think we're OK

Issue #497, opened 2026-05-21
Job_chat: Model outputs an empty or partial response


hanna-paasivirta:
This test scenario asking the model to explain codes fails in at least 40% of cases both via `job_chat` and `global_chat` on `main`. The model either emits an empty answer, three dots '...' or begins to explain code but ends the answer with a colon, with no code inline or in the `suggested_code` key.

I dug into this extensively but don't understand it. There are some known issues with using adaptive thinking and structured outputs, but there is probably something we are triggering inadvertently. It seems like there's too many constraints on the output and the model decides to stop the turn. 

I suspected that the ordering with code before text wasn't the most natural solution. I tried reversing it, and the problem does still appear (at least, cutting off on a colon: "...Here's an example that adds multiple members to your list in a single batch:", 'suggested_code': ""), but it might happen less frequently (1 in 10 tests).

I added the test in `services/global_chat/tests/acceptance/bugs/test_repro_dots_response.md` and `services/job_chat/tests/acceptance/bugs/test_repro_dots_response.md` in `global-chat-job-code` in #495 

```
hanna@Hannas-MacBook-Pro-2 apollo % poetry run pytest services/job_chat/tests/acceptance/tmp/test_repro_dots_response.md -s
=============================================================================================================== test session starts ================================================================================================================
platform darwin -- Python 3.11.15, pytest-8.4.1, pluggy-1.6.0
codspeed: 3.2.0 (disabled, mode: walltime, timer_resolution: 41.7ns)
benchmark: 5.1.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/hanna/openfn/apollo
configfile: pyproject.toml
plugins: recording-0.13.4, anyio-4.9.0, syrupy-4.9.1, socket-0.7.0, langsmith-0.4.1, codspeed-3.2.0, benchmark-5.1.0, asyncio-0.26.0
asyncio: mode=Mode.STRICT, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 1 item

services/job_chat/tests/acceptance/tmp/test_repro_dots_response.md
→ job-chat.tmp.repro-dots-response
  service: job_chat
  judges:  general, openfn_code_quality
  calling job_chat...
Calling services/job_chat ...

[...]

PR #528, opened 2026-06-15
Use Opus and environment variables for model selection

### LINKED ISSUE #533: Upgrade model to Opus
We switched to Fable, but it was taken down and so we've reverted to using Sonnet as our main user-facing chat model. Test use of Opus instead.

hanna-paasivirta:
## Short Description

Fable was retired, which broke every chat service that pointed at it. This moves the main chat model off Fable and makes it configurable. Model selection now lives in one place, and optional env vars let us change the live model without a redeploy.

Fixes https://github.com/OpenFn/apollo/issues/533 and https://github.com/OpenFn/apollo/issues/534

## Implementation Details

- `services/models.py` owns the whole model story: a default (Opus), a per-service map, and `preferred_chat_model(service)`.
- Each service resolves its model from its own env var if set, otherwise its code default, otherwise the global default. There is one env var per service, no catch-all.
- workflow_chat defaults to Sonnet, not Opus. It forces JSON/YAML output through structured outputs, and Opus handles that worse than Sonnet right now. The other services default to Opus.
- The service yamls no longer set a model. They point a comment at `models.py`.
- Optional env vars, one per agent: `APOLLO_GLOBAL_CHAT_MODEL` (planner), `APOLLO_WORKFLOW_CHAT_MODEL`, `APOLLO_JOB_CHAT_MODEL`. Unset by default. doc_agent has no var and runs on the default.
- Tests for models.py live in a new services/tests/unit/ directory, since models.py is a shared root module not owned by any one service.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,12 +10,12 @@ LANGFUSE_SECRET_KEY=sk-lf-...
 LANGFUSE_PUBLIC_KEY=pk-lf-...
 LANGFUSE_BASE_URL=https://cloud.langfuse.com
 
-# Chat model overrides (optional). Defaults live in services/models.py; set any
-# of these to switch the live model without redeploying. Accepts an alias
-# (claude-opus, claude-sonnet) or a full model ID.
-# APOLLO_CHAT_MODEL=          # global: forces all chat services to this model
-# APOLLO_WORKFLOW_CHAT_MODEL= # workflow_chat only (overrides the global var)
-# APOLLO_JOB_CHAT_MODEL=      # job_chat only (overrides the global var)
+# Chat model overrides (optional, one per service). Defaults live in
+# services/models.py; set a var to switch that service's live model without
+# redeploying. Accepts an alias (claude-opus, claude-sonnet) or a full model ID.
+# APOLLO_GLOBAL_CHAT_MODEL=   # global_chat planner
+# APOLLO_WORKFLOW_CHAT_MODEL= # workflow_chat
+# APOLLO_JOB_CHAT_MODEL=      # job_chat
 
 # HF_ACCESS_TOKEN=hf_YOUR-API-KEY-HERE # llama2 base
 # ZILLIZ_URI =  https://in01-XXXXXXXXXXXXX.aws-us-west-2.vectordb.zillizcloud.com:XXXXX
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,13 @@ LANGFUSE_SECRET_KEY=sk-lf-...
 LANGFUSE_PUBLIC_KEY=pk-lf-...
 LANGFUSE_BASE_URL=https://cloud.langfuse.com
 
+# Chat model overrides (optional). Defaults live in services/models.py; set any
+# of these to switch the live model without redeploying. Accepts an alias
+# (claude-opus, claude-sonnet) or a full model ID.
+# APOLLO_CHAT_MODEL=          # global: forces all chat services to this model
+# APOLLO_WORKFLOW_CHAT_MODEL= # workflow_chat only (overrides the global var)
+# APOLLO_JOB_CHAT_MODEL=      # job_chat only (overrides the global var)
+
 # HF_ACCESS_TOKEN=hf_YOUR-API-KEY-HERE # llama2 base
 # ZILLIZ_URI =  https://in01-XXXXXXXXXXXXX.aws-us-west-2.vectordb.zillizcloud.com:XXXXX
 # ZILLIZ_TOKEN =db_admin:password (or ApiKey)
--- a/services/doc_agent_chat/config.yaml
+++ b/services/doc_agent_chat/config.yaml
@@ -1,7 +1,6 @@
 config_version: 1.0
-# Chat model is set centrally in services/models.py (CHAT_MODEL_DEFAULT) and can
-# be overridden at runtime via the APOLLO_CHAT_MODEL env var. Add a `model:` key
-# here only to pin this service to a specific model.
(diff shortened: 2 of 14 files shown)


hanna-paasivirta:
I'm seeing the garbled model outputs triggered by structured outputs in `workflow_chat`. I've never seen these before in that service with Sonnet, but saw them in `job_chat` and fixed them by adding [a code edit tool](https://github.com/OpenFn/apollo/issues/497) and limiting the use of structured outputs to that only. In order to switch to Opus safely, I may need to [make a similar architecture change to `workflow_chat`](https://github.com/OpenFn/apollo/issues/530). To keep `workflow_chat` on Sonnet, I'll make the model setting more fine-grained and set it by service instead.

services/doc_agent_chat/config.yaml
```diff
@@ -1,5 +1,6 @@
 config_version: 1.0
-model: claude-fable
+# The chat model is configured in services/models.py (the default; doc_agent has
```
josephjclark, line 2: this comment doesn't make sense in isolation: it only makes sense if you know that the model used to be set in config. It's confusing and we should remove it. Same for the other models.

josephjclark: I don't actually love it 😬 Intuitively it feels that this config should be defaults for all values, and env vars can be used to  override it(somehow, it's not entirely practical!)

The split now of some things being envs and some things being config values feels confusing, rigid and arbitrary

services/models.py
```diff
@@ -3,22 +3,65 @@
...
+    "job_chat":      {"default": CLAUDE_OPUS,   "env": "APOLLO_JOB_CHAT_MODEL"},
+    "global_chat":   {"default": CLAUDE_OPUS,   "env": "APOLLO_GLOBAL_CHAT_MODEL"},
+}
+
+
+def preferred_chat_model(service: str | None = None) -> str:
+    """Resolve the main chat model for `service`.
+
+    Precedence: the service's env var if set, else its per-service default, else
+    CHAT_MODEL_DEFAULT. Each service's env var (e.g. APOLLO_WORKFLOW_CHAT_MODEL)
```
josephjclark, line 55: these comments are so so verbose. I think I need to start pushing back on them. The lightning codebase is probably more comment than code now.

Anyway this second sentence I don't like. It's repetitive, plus the "we can switch models without redeploying" thing is misleading. To change an env var you have to configure kubernetes and then restart the service.

It would be more accurate to say you can update it without a _rebuild_. But I wouldn't even say that at this level.

.env.example
```diff
@@ -10,6 +10,13 @@ LANGFUSE_SECRET_KEY=sk-lf-...
 LANGFUSE_PUBLIC_KEY=pk-lf-...
 LANGFUSE_BASE_URL=https://cloud.langfuse.com
 
+# Chat model overrides (optional, one per service). Defaults live in
+# services/models.py; set a var to switch that service's live model without
+# redeploying. Accepts an alias (claude-opus, claude-sonnet) or a full model ID.
+# APOLLO_GLOBAL_CHAT_MODEL=   # global_chat planner
+# APOLLO_WORKFLOW_CHAT_MODEL= # workflow_chat
+# APOLLO_JOB_CHAT_MODEL=      # job_chat
```
josephjclark, line 18: These sample env vars don't make sense do that? What does `job_chat` resolve to?

josephjclark (changes requested):
Hi @hanna-paasivirta - I just wanted to post some initial impressions. Unfortunately this project really doesn't sit well with me, and seeing the implementation makes me very nervous.

I'm only half-way through the review and need to give it some more time, but I wanted to post where I'd got to so far.

Also - despite very heavy AI commenting there's no readme documentation about how this works (just some very spurious looking env var defaults?). We should correct that because the relationship between envs and config is murky

josephjclark:
@hanna-paasivirta what I think makes more sense here is:

a) to hard-code each service to a particular model, as we do on main
b) to use env vars to drive the version of each model

So you'd have an env var `OPUS_VERSION` with a value of `4-8`.

Then `models.py` would have a function like `getModelVersion(name)` or something. And job_chat  calls `getModelVersion('opus')`, which returns `claude-opus-4-8` (where the version suffix comes from env or a default).

The model name can still come from the config.yaml file for each service, which is where any service specific stuff lives. But it would only have a model name, not a version.

Basically this means that the env only drives the version number, not the model itself. Otherwise the code is much as it is on main right now, where the service itself makes the big decisions about which model to use, and the env just bumps the version to keep it modern.

As I've said on slack: this architecture would not have helped us with the fable switch-off: fable support needed more than just a version string, and a dynamic downgrade likely needs more to. If we want to be robust to models disappearing overnight (a very worrying precedent) we should put some thought into be better rollback solutions (I'm aware there's another PR open for this)

hanna-paasivirta:
> As I've said on slack: this architecture would not have helped us with the fable switch-off: fable support needed more than just a version string, and a dynamic downgrade likely needs more to. If we want to be robust to models disappearing overnight (a very worrying precedent) we should put some thought into be better rollback solutions (I'm aware there's another PR open for this)

We needed more because we were upgrading the model to a new one. The added protections work ok for existing models. But there's little guarantee that we can always downgrade the model easily in the future like we can now with fable/opus/sonnet

hanna-paasivirta:
> Basically this means that the env only drives the version number, not the model itself. 

I think there has never been a situation where the pointer to a model version (without specifying a snapshot) would stop work working and require an intervention on our side. But to be fair a model family had never been taken down before either.

PR #529, opened 2026-06-15
Add a model fallback chain

hanna-paasivirta:
## Short Description

If a chat service's model is unavailable, it now falls back to another model instead of failing the request. Each fallback is reported to Sentry, because it is meant to be temporary.

Closes https://github.com/OpenFn/apollo/issues/535

## Implementation Details

- Two small wrappers in `services/models.py` wrap each chat call, one for normal calls and one for streaming.
- The chain is Opus, then the previous Opus, then Sonnet, starting from the service's preferred model.
- Fallback triggers when a model is removed (404) or the provider is down or overloaded (500, 502, 503, 529). The SDK already retries transient errors before this point.
- Streaming only falls back before the first token reaches the user. After that an error surfaces, so we never re-stream a duplicate answer.
- Normal errors like bad request, auth, and rate limit are not fallbacks and surface immediately.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/services/doc_agent_chat/agent.py
+++ b/services/doc_agent_chat/agent.py
@@ -7,7 +7,7 @@ from doc_agent_chat.doc_search import DocSearch
 from doc_agent_chat.prompt import build_system_prompt
 from doc_agent_chat.tools import TOOL_DEFINITIONS, search_documents, format_search_results_as_documents
 from doc_agent_chat.config_loader import ConfigLoader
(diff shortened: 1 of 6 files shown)


hanna-paasivirta:
Editing this to set models per service then retest.  See https://github.com/OpenFn/apollo/pull/528#issuecomment-4710526508

PR #538, opened 2026-06-18
Add authentication

### LINKED ISSUE #545: Epic: Authentication MVP
# Apollo can't tell its callers apart or control the keys they use

## Problem / Context

Apollo serves AI features to OpenFn deployments, but it has no way to recognise who is calling. Today a caller has to bring their own Anthropic key on each request, which Apollo simply forwards to the LLM.

That means Apollo can't associate a request with a particular OpenFn instance, can't hold and use a key on a client's behalf, and can't control which callers it trusts.

As more self-hosted Lightning instances start pointing at our hosted Apollo, separate deployments increasingly share one production service, and this gap matters.

## Goal

When this is done, Apollo can recognise a known OpenFn client from the request it already sends, and use the AI provider key we hold for that client rather than relying on the caller to supply one.

Callers we don't recognise carry on exactly as they do today. There is one shared, production Apollo, so introducing this must not break anything already running against it.

## Scope

**In:**

- A way to register an OpenFn client with Apollo and associate an AI provider key with it.
- Recognising a known client on an incoming request and using its associated key.
- Storing client credentials and provider keys safely at rest.

**Out (for now):**

- Forcing every caller to be a known client / turning Apollo into a known-clients-only service.
- Any change on the OpenFn (Lightning) side to make this work.
- Per-endpoint access tiers or fine-grained authorisation.
- Richer error reporting that distinguishes "we don't know you" from "your provider key failed" (wanted, but a separate follow-on).

## Acceptance

- A known client's request is served using the key we hold for them, without the caller having to supply one.
- A caller we don't recognise is served exactly as it is on `main` today — no new rejections, across every transport and route that works now.
- Stored secrets are not readable in the clear if the store leaks.
- Apollo's own internal service-to-service calls keep working.
- Rolling this out onto the shared production instance changes nothing for the OpenFn instances already using it.

## Follow-ups

- [ ] **WebSocket identity.** Change the WebSocket handshake so a known client can identify itself on that transport. Today it can't carry the credential the way an HTTP request body does, so client recognition only works on the request/response and streaming paths.
- [ ] **Provision OpenFn's own key.** Register OpenFn itself as a client and give it a held provider key, so OpenFn's own traffic uses a held key rather than a forwarded one.

taylordowns2000:
Adds a `lightning_clients` table so that Apollo can:

1. authenticate lightning instances who attempt to use its services
2. handle the issuance of ai service keys (e.g., anthropic API keys) and associate those keys with specific clients 

It is designed as a MVP, and built to be opt-in.

 closes #545 

## To migrate

1. add `INSTANCE_AUTH=true` to your `.env`
2. add `APOLLO_ENC_KEY` to your `.env` (or secret store) with `openssl rand -base64 32`
3. add the `lightning_clients` table with `set -a; . ./.env; set +a; psql "$POSTGRES_URL" -f services/_instance_auth/schema.sql`
4. provision a token for any lightning instance you want to connect with `bun services/_instance_auth/provision_client.ts <some-identifier-123> <some-sk-ant-token-blah-456>`
5. Add this hashed/encrypted record to the Apollo DB
6. Add this un-hashed `AI_ASSISTANT_API_KEY` (i.e., the secret!) to your lightning `.env` (or secret store)
7. Repeat for as many instances as you'd like to connect.
8. Optionally, merge this PR so that Lightning users see better error messages if their tokens are rejected by Apollo and/or downstream AI providers: https://github.com/OpenFn/lightning/pull/4879 

## _After_ this PR (what's next)

Related to @elias-ba 's concurrent proposal to enhance error handling on Lightning, my **next goal** is to let Lightning reliably tell **"we rejected your api_key"** from **"your api_key is fine, but the Anthropic key we use for you failed / is over its limit."** HTTP status collides (both are `401`), so the discriminator is the **`type` field**, not the status code.

### Proposed contract (`type` is the discriminator)

| `type` | HTTP | Meaning to client | Source |
|---|---|---|---|
| `UNAUTHORIZED` | 401 | Lightning api_key not recognised | The gate, **only** the gate |
| `AUTH_ERROR` | 401 | api_key fine; Anthropic key invalid/revoked | Anthropic/LiteLLM `AuthenticationError` |
| `RATE_LIMIT` | 429 | Anthropic key hit a rate/usage limit, retry later (`retry_after`) | Anthropic/LiteLLM `RateLimitError` |
| `BILLING` *(new)* | 402 | Anthropic key out of credit / over spend cap | `billing_error` type + "credit balance too low" |
| `KEY_UNAVAILABLE` *(new)* | 503 | Server-side: stored key can't be decrypted (`APOLLO_ENC_KEY` misconfig) | `auth.ts`, not the client's fault |

Rule: `UNAUTHORIZED` is reserved for credential rejection. Anything touching the Anthropic/LiteLLM key returns `AUTH_ERROR` / `RATE_LIMIT` / `BILLING`, never `UNAUTHORIZED`.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/platform/src/middleware/auth.ts
+++ b/platform/src/middleware/auth.ts
@@ -3,64 +3,44 @@ import { createHash, randomBytes, timingSafeEqual } from "node:crypto";
 import type { ApolloError } from "../util/errors";
 import { ENC_PREFIX, decryptKey, parseEncKey } from "../util/instance-key-crypto";
 
-// A row from the lightning_clients table, as used at runtime.
 type Client = { name: string; anthropicKey: string | null };
-
-// A lookup resolves a token hash to a client, or null if unknown.
 type Lookup = (hash: string) => Promise<Client | null> | Client | null;
 
-// Instance auth is opt-in via the INSTANCE_AUTH env var (see initAuth). When it
-// is unset/falsey, /services/* is open exactly as before. When set, every
-// /services/* request must carry an api_key (in the body) that matches a known
-// client.
+// Opt-in via INSTANCE_AUTH (see initAuth); when off, /services/* is open as before.
 let enabled = false;
 
-// True once the lightning_clients lookup is actually usable. When auth is
-// enabled but the DB/table can't be reached, this stays false and the gate
-// fails CLOSED (rejects every external caller) rather than silently opening up.
+// False until the lightning_clients lookup is usable. Auth-on but unusable DB =>
+// the gate fails CLOSED (rejects every external caller) rather than opening up.
 let dbReady = false;
 
-// When set (by tests), this replaces the DB-backed lookup so the suite can
-// enable auth with a known token set without a real Postgres.
 let lookupOverride: Lookup | null = null;
-
-// Bun's native Postgres client (Bun >= 1.2). Created once in initAuth().
 let sql: SQL | null = null;
 
-// In-memory cache of all clients keyed by token hash, refreshed on a TTL so
-// rows added/revoked directly in the DB are picked up within CACHE_TTL_MS.
+// In-memory client cache (keyed by token hash), refreshed on a TTL so DB changes
+// are picked up within CACHE_TTL_MS.
 const CACHE_TTL_MS = 60_000;
 let clientCache: Map<string, Client> | null = null;
 let cacheTs = 0;
 
-// Single-flight handle for the cache refresh. Concurrent refreshes (e.g. a burst
-// of requests hitting the TTL boundary at once) all share THIS one promise, so a
-// refresh can never trigger more than one DB read no matter the concurrency.
+// Single-flight handle: concurrent refreshes share this one promise, so a refresh
+// can never trigger more than one DB read no matter the concurrency.
 let refreshInFlight: Promise<Map<string, Client>> | null = null;
 
-// Master key (32 bytes) for decrypting stored anthropic_api_key values, parsed
-// from APOLLO_ENC_KEY in initAuth. Null when unset — plaintext rows still work,
-// but any "enc:v1:" row can't be decrypted and its client is omitted.
+// Master key for decrypting stored anthropic_api_key values (APOLLO_ENC_KEY). Null
+// when unset — plaintext rows still work, but "enc:v1:" rows can't be decrypted.
 let encKey: Buffer | null = null;
 
-// When set (by tests), this replaces loadClients in the refresh path so the
-// single-flight/stale-while-revalidate behaviour can be exercised without a DB.
 let loaderOverride: (() => Promise<Map<string, Client>>) | null = null;
 
-// A per-process secret that identifies genuine Apollo-to-Apollo calls. The
-// bridge spawns Python services as child processes that inherit this process's
-// env, so exporting the token here lets services/util.py apollo() echo it back
-// via the X-Apollo-Internal header. authGate exempts any request carrying it,
-// so "Apollo calling itself" skips auth without trusting network position (the
-// old loopback exemption trusted any local caller, including Lightning).
-// Honour an operator-provided value; otherwise mint a fresh random one.
+// Per-process secret identifying genuine Apollo-to-Apollo calls. The bridge spawns
+// Python children that inherit this env, so services/util.py apollo() echoes it back
+// via the internal header; authGate exempts requests carrying it without trusting
+// network position. Honour an operator-provided value, else mint a random one.
 //
-// MULTI-PROCESS DEPLOYMENTS: when unset, each process mints its OWN random
-// token. apollo() self-calls target 127.0.0.1:{port}, so they normally hit the
-// same process and the minted token matches. But if several processes share a
-// port (e.g. SO_REUSEPORT / clustering), a self-call can land on a sibling
-// whose token differs and get a spurious 401. Set APOLLO_INTERNAL_TOKEN to the
-// SAME value across all processes in that case so they recognise each other.
+// MULTI-PROCESS: when unset, each process mints its OWN token. apollo() self-calls
+// hit 127.0.0.1:{port} and normally land on the same process, but if processes
+// share a port (SO_REUSEPORT / clustering) a self-call can hit a sibling and 401.
+// Set APOLLO_INTERNAL_TOKEN to the SAME value across processes in that case.
 export const INTERNAL_HEADER = "x-apollo-internal";
 const INTERNAL_TOKEN =
   process.env.APOLLO_INTERNAL_TOKEN ?? randomBytes(32).toString("hex");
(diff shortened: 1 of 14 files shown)


platform/src/middleware/auth.ts
```diff
@@ -0,0 +1,308 @@
...
+      "Apollo instance auth ENABLED but POSTGRES_URL is not set — clients cannot be looked up. /services/* will REJECT all external callers until this is fixed."
+    );
+    return;
+  }
+
+  try {
+    sql = new SQL(url);
+    const rows = (await sql`
+      SELECT to_regclass('public.lightning_clients') AS t
+    `) as Array<{ t: string | null }>;
```
elias-ba, line 202: This is the path breaks on breaks on any environment respecting the pinned Bun version in `.tools-version`: `SQL` is undefined on <1.2.0, so `new SQL(url)` throws here. It's swallowed by the catch below, so there's no crash - just a silent fail-closed gate and a log line. The whole real-DB path (this + loadClients) has no CI coverage because every test goes through the `__setLoaderForTest` seam. Worth one integration test that exercises `initAuth`/`loadClients` against a real PG so this can't rot again.

platform/src/middleware/auth.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  // The credential is the api_key the caller sends in the body; hash it and look
+  // for a matching client. Absent or unknown -> 401.
+  const apiKey =
+    typeof ctx.body?.api_key === "string" ? ctx.body.api_key.trim() : "";
+  if (!apiKey) return unauthorized(ctx);
+
+  const lookup = lookupOverride ?? dbLookup;
+  const client = await lookup(hashToken(apiKey));
+  if (!client) return unauthorized(ctx);
```
elias-ba, line 255: Semantics changed from #536: that PR *recognized* tokens and swapped keys but passed unknown callers through untouched, so several instances could share one Apollo and the un-provisioned ones kept working as today. Here, any unknown or missing api_key is a hard 401 when enabled - it's all-or-nothing per Apollo process. That's correct for a dedicated Apollo instance, but flipping `INSTANCE_AUTH` on our shared production Apollo would 401 every not-yet-provisioned instance. Worth really noting and paying attention to. Can you confirm this is what we really want ?

services/_instance_auth/hash_token.py
```diff
@@ -0,0 +1,47 @@
...
+platform/src/middleware/auth.ts (sha256 over UTF-8 bytes, lowercase hex).
+"""
+
+import hashlib
+import secrets
+import sys
+
+
+def hash_token(token: str) -> str:
+    return hashlib.sha256(token.encode()).hexdigest()
```
elias-ba, line 19: The docstring says this "must match the hash Apollo computes in auth.ts" - but #538 dropped the pinned-hash test that enforced it (#536 had KNOWN_TOKEN/KNOWN_HASH asserted on both the TS and Python sides). With three independent SHA-256 sites now (this, auth.ts:62, provision_client.ts:34) and nothing pinning them, a drift would silently 401 every provisioned client - the exact "nightmare to debug" failure mode. Please restore the cross-language pinned test.

services/_instance_auth/provision_client.ts
```diff
@@ -0,0 +1,54 @@
...
+      "Run this from the repo root so Bun picks up .env. If it's still missing,\n" +
+      "the key isn't in .env yet — add one with:\n\n" +
+      '  echo "APOLLO_ENC_KEY=$(openssl rand -base64 32)" >> .env\n\n' +
+      "then re-run (and restart Apollo so it can decrypt at runtime)."
+  );
+  process.exit(1);
+}
+
+const apiKey = randomBytes(32).toString("base64url");
+const authTokenHash = createHash("sha256").update(apiKey).digest("hex");
```
elias-ba, line 34: Re-implements the hash inline instead of importing `hashToken` from auth.ts. Since this is the canonical provisioning path, a drift here vs. the runtime hasher silently locks clients out. Import it instead:

```javascript
import { hashToken } from "../../platform/src/middleware/auth";
...
const authTokenHash = hashToken(apiKey);
```

platform/src/middleware/auth.ts
```diff
@@ -0,0 +1,308 @@
...
+// network position. Honour an operator-provided value, else mint a random one.
+//
+// MULTI-PROCESS: when unset, each process mints its OWN token. apollo() self-calls
+// hit 127.0.0.1:{port} and normally land on the same process, but if processes
+// share a port (SO_REUSEPORT / clustering) a self-call can hit a sibling and 401.
+// Set APOLLO_INTERNAL_TOKEN to the SAME value across processes in that case.
+export const INTERNAL_HEADER = "x-apollo-internal";
+const INTERNAL_TOKEN =
+  process.env.APOLLO_INTERNAL_TOKEN ?? randomBytes(32).toString("hex");
+process.env.APOLLO_INTERNAL_TOKEN = INTERNAL_TOKEN;
```
elias-ba, line 47: Multi-process caveat is documented here - just flagging for the deploy: if prod runs Apollo with SO_REUSEPORT/clustering and APOLLO_INTERNAL_TOKEN unset, apollo() self-calls can hit a sibling process with a different token and 401. Confirm our topology is single-process, or set a shared APOLLO_INTERNAL_TOKEN.

platform/src/middleware/auth.ts
```diff
@@ -0,0 +1,308 @@
...
+// False until the lightning_clients lookup is usable. Auth-on but unusable DB =>
+// the gate fails CLOSED (rejects every external caller) rather than opening up.
+let dbReady = false;
+
+let lookupOverride: Lookup | null = null;
+let sql: SQL | null = null;
+
+// In-memory client cache (keyed by token hash), refreshed on a TTL so DB changes
+// are picked up within CACHE_TTL_MS.
+const CACHE_TTL_MS = 60_000;
```
elias-ba, line 21: Confirming this is an accepted trade-off: a revoked/rotated credential keeps working for up to ~60s (until the cache TTL refreshes). Fine for rotation; just want it acknowledged for the security posture, since "revoke" isn't immediate without a restart.

elias-ba (changes requested):
Hey @taylordowns2000, this is genuinely strong work and a clever evolution of the concept. Using the existing `api_key` field as the credential is the right call, it gets us per-client keys with no Lightning code change. Encryption fails closed correctly, the single-flight + stale-while-revalidate cache is solid and well tested, and swapping the loopback exemption for a per-process internal token is a real security improvement. Two blockers before we enable this, then a couple of things I'd like you to confirm.

**Blockers**

1. `.tool-versions` still pins `bun 1.1.13`, but `auth.ts` imports `{ SQL } from "bun"`, which doesn't exist before Bun 1.2.0 (I checked, `typeof SQL` is `undefined` on the current toolchain). With `INSTANCE_AUTH=true`, `new SQL(url)` throws, gets caught, and the gate fails closed, so every external caller gets a 401. #536 bumped this to `bun 1.3.14` for exactly this reason and it needs to come along here.

2. The pinned cross-language hash test from #536 got dropped. We now have three SHA-256 sites with nothing enforcing they agree, and any drift silently 401s every provisioned client. Please restore it. (inline)

**Needs confirmation**

- The gate is all-or-nothing per Apollo process, so turning `INSTANCE_AUTH` on for our shared production Apollo would 401 every instance we haven't provisioned yet. Totally fine for dedicated Apollo instances. But concerning for shared ones. I just want to confirm that's the plan. Also worth being precise that "no Lightning-side change" is true for code, but each instance's configured key value still has to be swapped to the minted credential.

- WS transport is fully disabled while auth is on, since the upgrade has no body to carry the key. Should be fine since Lightning uses POST and /stream, but can you confirm nothing still speaks WS?

- No changeset entry on this one and we've been keeping those (#536 had one), worth adding before release.

platform/src/middleware/auth.ts
taylordowns2000: Yes, intentional. I don't think there's a benefit in hard cutoff.

platform/src/middleware/auth.ts
taylordowns2000: Yes, absolutely. That's the idea of the initiative.

platform/test/server.test.ts
```diff
@@ -124,9 +141,271 @@ describe("Python Services", () => {
...
+
+    it("rejects a missing api_key with 401 UNAUTHORIZED", async () => {
+      const res = await app.handle(post("services/echo", { x: 1 }));
+      expect(res.status).toBe(401);
+      const body = await res.json();
+      expect(body.code).toBe(401);
+      expect(body.type).toBe("UNAUTHORIZED");
+    });
+
+    it("rejects an unknown api_key with 401", async () => {
```
josephjclark, line 204: This breaks production

platform/src/middleware/auth.ts
elias-ba: Too big and too restrictive for now. The approach I would take and that I proposed in #536 was to unblock the current with the minimum and we discuss this and build it in another phase.

platform/test/server.test.ts
taylordowns2000: That's correct! Did you catch up with @stuartc about this? Migration notes are in the PR

PR #540, opened 2026-06-18
fix: job_chat drops generated code when the job is empty

### LINKED ISSUE #539: Apollo fails when an empty job string is sent
If you send a chat to Apollo, asking it to write a simple http post but you have _nothing_ (nope, not even a comment) in your job body, it will think it's capable of helping you out but actually fall flat on its face!

https://www.loom.com/share/c2e520a2c2af434e8fcd834cf40deb26

This is super frustrating, since it suggests that it's doing something for you, but never does. Here's how to repro:
1. create new workflow from "scheduled" template
2. open first step (click step, press `command+e`)
3. remove all text from job code (select all, backspace)
4. send chat to apollo: "write a simple http post for me"
5. see it insist that it's done so, but fail to put anything in your code editor.
6. === the workaround ⬇️ ===
7. add some text to you code editor
8. send the same message again
9. watch it work

Failing on `v2.16.7 (1.5.87)` with `apollo:v1.3.0` (but likely not addressed in `1.3.1` or `1.3.2` either.)

elias-ba:
## Short Description

`job_chat` generated code but returned nothing when the job body was empty. This makes it return the code.

Fixes #539

## Implementation Details

`job_chat` writes job code by editing what's already there. When the job is empty there's nothing to edit, so it dropped the code it generated and only returned the chat reply. Downstream that shows up as the assistant promising code and never delivering it.

Changes in `services/job_chat`:

- `generate()`: apply the edit even when the job is empty (a blank job is just a blank file to write into).
- `apply_single_edit()`: if the job is empty, use the new code directly no matter which edit action the model picked. Keeps it correct even when the model doesn't choose "rewrite", and skips a wasted retry.
- prompt: nudge the model to use "rewrite" on an empty job.

Added unit tests for the empty job cases (rewrite, replace, whitespace-only, and no edit).

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.changeset/ninety-meteors-shake.md
+++ b/.changeset/ninety-meteors-shake.md
@@ -1,0 +1,6 @@
+---
+"apollo": patch
+---
+
+Fix an issue in job_chat where no code is returned if an an empty job expression
+is sent
--- a/.changeset/ninety-meteors-shake.md
+++ b/.changeset/ninety-meteors-shake.md
@@ -1,6 +1,0 @@
----
-"apollo": patch
----
-
-Fix an issue in job_chat where no code is returned if an an empty job expression
-is sent
--- a/.changeset/ripe-poets-enjoy.md
+++ b/.changeset/ripe-poets-enjoy.md
@@ -1,5 +1,0 @@
----
-"apollo": patch
----
-
-update dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # apollo
 
+## 1.3.3
+
+### Patch Changes
+
+- 824b99f: Fix an issue in job_chat where no code is returned if an an empty job
+  expression is sent
+- e5daf45: update dependencies
+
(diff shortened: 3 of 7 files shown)


poetry.toml
```diff
@@ -0,0 +1,2 @@
+[virtualenvs]
+in-project = true
```
josephjclark: @elias-ba do you know what this file is?

josephjclark:
@elias-ba please run `bun changeset` to generate a changeset ready for release

taylordowns2000 (commented):
rejected from me, but only because @elias-ba and i cant get it to work on my machine. (workflow chat is OK, job chat is broken.) let's assume it's an issue with my local setup. will wait for someone else to review

josephjclark (approved):
Works great for me and the fix looks quite reasonable.

I've just pushed a simple changeset and I'm ready and willing to release

PR #547, opened 2026-06-23
Global assistant: send a full workflow YAML only for multistep changes

### LINKED ISSUE #507: Global chat: Applying job code changes across several steps
# Applying job code changes
In #490, we addressed job code edit scenarios involving a single job code step. In this issue, we'll focus on more complex scenarios involving more than one step, or more than one agent.

## Test scenarios
- User asks for a change to the current step that requires information from other steps ("make this step like my last step", "do the same elsewhere", "check for x in all steps")
- User asks for a change to a different step from another step/workflow view
- User asks for several changes across steps from one step/workflow view
- User asks to "add a step in my workflow to...[do something vague]" -> ok to call workflow agent directly and leave the code empty
- User asks to "add a step in my workflow to...[do something specific]" -> to fill in the code, we need to call the planner, workflow subagent and the job code subagent. 

## Apollo architecture work:
- Make sure the planner is able to and encouraged to examine any relevant parts of the full YAML with the YAML tool
- Verify that the internal YAML is kept up to date between tool calls, especially concurrent tool calls.
- Adjust payload to return the full YAML only, not YAML + a single changed step (see below)

## Required changes in Lightning:
- Essential for this to function: Currently, we can't implement changes in several steps from the job code view, as Lightning expects to receive a single job code change. We should update Lightning to consume YAML-only updates so the job-code agent can return multi-step changes in full and without mixing up steps.
- Essential for UX, but not for an experimental feature: A way to signal to the user from any workflow view/job code view which steps were changed.

hanna-paasivirta:
## DO NOT MERGE

Until  https://github.com/OpenFn/lightning/pull/4888 is in lightning prod

## Short Description

This changes the Global Assistant payload to always return a single full workflow YAML, instead of a YAML plus separate job code. This lets one conversation turn apply code changes across several steps.

This is a breaking change in the Global Assistant. It must be coordinated with Lightning: OpenFn/lightning#4888.

The PR also bundles a series of routing, tool, and prompt improvements to the Global Assistant.

Fixes #507.

## Payload

- Responses now carry one `workflow_yaml` attachment with the full workflow, with any generated job code already stitched in. There is no separate job-code artifact.
- A single source of truth lets the planner change structure and several job bodies in the same turn, then return one coherent YAML.

## Planner

- Adaptor selection is delegated to the workflow subagent. The planner describes which systems to connect and no longer names adaptors. It had no adaptor list and was picking the wrong ones.
- "Add a step" requests are split by intent. A vague request adds the step only and asks what it should do. A specific request also fills in the code via the job subagent.
- Added self-correction. If a subagent reply shows the wrong tool was used (for example it cannot see job code), the planner re-routes instead of passing the message to the user.
- Returned history now uses plain string content, matching the direct routes, instead of raw tool-call blocks. This keeps all three agents consistent and removes a latent replay crash.
- Raised the max tool calls from 10 to 25 so larger workflows are not cut off mid-build.

## Router

- Clarified the subagent roles. Structural edits go to the workflow agent, code goes to the job agent, and anything needing both, multiple steps, or another step as a reference goes to the planner. The job agent only ever sees one step, so requests that read or touch other steps must use the planner.
- "What does my workflow do" now routes on whether the steps contain code. It goes to the planner when they do, so the real code can be read, and to the workflow agent when they are empty, which is faster.

## Docs and tests

- Updated the global_chat README and payload spec to match the new behaviour.
- Added unit tests for the planner, the router, and the empty-job fix.
- Added acceptance tests for the user scenarios involving job code changes across several steps #507 

## AI Usage

- [x] Strategy / design
- [x] Other
- [ ] I have not used AI

### DIFF
--- a/.changeset/early-carrots-look.md
+++ b/.changeset/early-carrots-look.md
@@ -1,0 +1,5 @@
+---
+"apollo": major
+---
+
+enable global assistant multi step changes
--- a/poetry.lock
+++ b/poetry.lock
@@ -2164,14 +2164,14 @@ typing-extensions = ">=4.14.1"
 
 [[package]]
 name = "pydantic-settings"
-version = "2.14.0"
+version = "2.14.2"
 description = "Settings management using Pydantic"
 optional = false
 python-versions = ">=3.10"
 groups = ["main"]
 files = [
-    {file = "pydantic_settings-2.14.0-py3-none-any.whl", hash = "sha256:fc8d5d692eb7092e43c8647c1c35a3ecd00e040fcf02ed86f4cb5458ca62182e"},
-    {file = "pydantic_settings-2.14.0.tar.gz", hash = "sha256:24285fd4b0e0c06507dd9fdfd331ee23794305352aaec8fc4eb92d4047aeb67d"},
+    {file = "pydantic_settings-2.14.2-py3-none-any.whl", hash = "sha256:a20c97b37910b6550d5ea50fbcc2d4187defe58cd57070b73863d069419c9440"},
+    {file = "pydantic_settings-2.14.2.tar.gz", hash = "sha256:c19dd64b19097f1de80184f0cc7b0272a13ae6e170cbf240a3e27e381ed14a5f"},
 ]
 
 [package.dependencies]
--- a/poetry.lock
+++ b/poetry.lock
@@ -1146,14 +1146,14 @@ wrapt = ">=1.14,<2"
 
 [[package]]
 name = "langsmith"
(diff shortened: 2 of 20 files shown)


hanna-paasivirta:
@josephjclark Let us know if you want @elias-ba to review this

josephjclark:
Yes, this is a breaking change and is incompatible with older lightning versions.

But we can release this as soon as https://github.com/OpenFn/lightning/pull/4888 is live

PR #550, opened 2026-06-24
Auth: recognise known OpenFn clients and use the keys we hold for them

### LINKED ISSUE #545: Epic: Authentication MVP

stuartc:
## What & why

Apollo had no way to tell its callers apart. Every request brought its own Anthropic key, which we forwarded straight to the LLM, so we couldn't associate a request with a particular OpenFn instance or hold a key on a client's behalf. As self-hosted Lightning instances start pointing at our shared hosted Apollo, that matters.

This adds per-client recognition: a known client's request is served with the key we hold for them, while callers we don't recognise carry on exactly as they do on `main` today. No Lightning-side change is required.

Fixes #545

## How it works

We have introduced an "auth token", mapping a hashed key that can be provisioned and mapped against an Anthropic API key.

An always-on authenticate hook on `/services/*` looks up the `api_key` already in the request body by its SHA-256 in a `lightning_clients` table:

- **Known client** → the inbound key is never forwarded to the LLM. It's swapped for the client's stored `anthropic_api_key`, or stripped to fall back to the global key when that column is null.
- **Unknown key** → forwarded only if it's `sk-ant-`-shaped (bring-your-own); an unknown non-`sk-ant-` key is rejected with 401 so a Lightning credential can't leak to the LLM.
- **Internal Apollo-to-Apollo calls** → exempt via a per-process token.

Auth tokens are encrypted at rest using AES-256-GCM and cached in memory (current for ~60s).

Changes to the key (either key existence or DB availability) are reflected after expiry.

In production environments it's advised that the clients schema is kept separately, so there is a second (optional) env variable, `APOLLO_CLIENTS_DB_URL`, which falls back to `POSTGRES_URL` when absent.

There is a CLI script to help with auth tokens and keys via `bun run client`. API tokens are added and rotated via stdin, to prevent tokens being logged or stored in shell history. Bear in mind that since auth tokens are hashed, once they've been printed in the terminal they are unrecoverable.

## Commits

Six logical commits, readable in order — Bun/lockfile chore, DB pool + migration runner, the auth hook, the provisioning CLI, tests, then docs.

**Other changes**

Both CI and the Docker image build use the version of Bun declared in `.tool-versions`, and Bun is now pinned to the current latest (1.3.14).

## Known Issues

The WebSockets endpoint doesn't work right now, because WS handshakes can't carry the body credential. This will be sorted in a follow-up.


josephjclark (approved):

README.md
```diff
@@ -189,7 +208,143 @@ Python scripts are invoked through a child process. Each call to a service runs
 in its own context.
 
 Python modules are pretty free-form but must adhere to a minimal structure. See
-the Contribution Guide for details.
+the [Contribution Guide](CONTRIBUTING.md) for details.
+
+## Services
```
josephjclark, line 213: This is an intrusive and annoying change which adds little value. I don't want to have to maintain this each time we edit services.

Services should be self documenting. Plus the server homepage lists out all services. I'd rather guide users to that.

I'm going to remove this from the next PR

PR #558, opened 2026-06-26
Version stuff

josephjclark:
## Short Description

Adds a bunch of tracking and versioning utilities to apollo

- Adds a server version number as a HTTP header on all responses; on chat payloads under the `meta` key; and on the langfuse client
- ~For traces with code attached, add a `has_code_attachment` tag in langfuse~

Why?

- Generally I've been thinking for a while now that apollo should report its version better
- The langfuse tag will improve the quality of our langfuse evals (Angus will thank me for it)
- I want to use the version number in chat payloads to guard against an upcoming breaking change

## AI Usage

Please disclose whether you've used AI in this work (it's cool, we just want to know!):

- [x] Yes, I have not used AI
- [ ] No, I have not used AI

### DIFF
--- a/platform/src/server.ts
+++ b/platform/src/server.ts
@@ -11,6 +11,7 @@ import { captureException } from "./util/sentry";
 import { clientsDbUrl, closeDb } from "./db";
 import { runMigrations } from "./db/migrate";
 import { randomUUID } from "node:crypto";
+import pkg from "../../package.json";
 
 export default async (
   port: number | string = 3000,
@@ -23,6 +24,7 @@ export default async (
   app.use(html());
 
   app.derive(() => ({ start: Date.now(), uuid: randomUUID() }));
+  app.onAfterHandle(({ set }) => { set.headers["X-Api-Version"] = pkg.version; });
   app.onAfterHandle(logRequest);
 
   // Report unhandled throws to Sentry, then return nothing so Elysia produces
--- a/.changeset/honest-canyons-cough.md
+++ b/.changeset/honest-canyons-cough.md
@@ -1,0 +1,6 @@
+---
+"apollo": patch
+---
+
+Add version metadata to HTTP headers, chat payloads (meta key), and langfuse
+traces
--- a/.changeset/two-flowers-lay.md
+++ b/.changeset/two-flowers-lay.md
@@ -1,0 +1,5 @@
+---
+"apollo": patch
+---
+
+For langfuse traces with code, add a `has_code_attachment` tag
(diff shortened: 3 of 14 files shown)


platform/src/server.ts
```diff
@@ -23,6 +24,7 @@ export default async (
   app.use(html());
 
   app.derive(() => ({ start: Date.now(), uuid: randomUUID() }));
+  app.onAfterHandle(({ set }) => { set.headers["X-Api-Version"] = pkg.version; });
```
elias-ba, line 27: I think `onAfterHandle` only fires on the success path, so I'm not sure the header actually lands on all responses like the description says. Errors go through `onError` which doesn't set it, so wouldn't 401s from the auth hook and `ApolloError`/500s come back without the version? And for the streamed `/services/*/stream` responses, aren't the headers already flushed by the time this runs?

Not a blocker at all since the breaking-change guard rides on `meta.apollo_version` anyway. But should we set it somewhere that also covers the error/stream paths (mirror it in `onError`, or move it to `onRequest`/`mapResponse`)? And maybe add a test for an error response too, since the current one only checks the happy path?

elias-ba (approved):
Hey @josephjclark, had a look at this and it's really nice work. I've left one inline comment/question - totally up to you whether it's worth handling now or we just merge as is. Either way this is great stuff, well done and thanks! 🙌

platform/src/server.ts
josephjclark: Great catch - thanks I'll take a close look at this

josephjclark: So I've just added a 400 response from python and that includes the header
```
$ curl http://localhost:3000/services/echo  -v -X POST
* Host localhost:3000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:3000...
* Connected to localhost (::1) port 3000
> POST /services/echo HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< X-Api-Version: 1.4.0
< Date: Fri, 26 Jun 2026 14:00:29 GMT
< Content-Length: 59
```

But in this case the typescript handler returns successfully. So if the handler threw an internal error for some reason we might not get the version number.

You make a good point about streaming though. I'm not actually sure where we'd return the version for a streaming API. I think I'll raise an issue for this.

One thing to note is that Lightning will make a HEAD call to the endpoint to get the version before it does anything. So the streaming thing is academic

josephjclark:
Ach! Turns out the tag stuff doesn't work at all. I'm pulling this and we'll have to look at it later.

PR #564, opened 2026-06-29
Remove anthropic api token from http auth

stuartc:
## Short Description

Tightens instance auth so an inbound `api_key` is only ever treated as a client credential, never forwarded to Anthropic. Drops the old `sk-ant-` bring-your-own-key path.

Fixes #

## Implementation Details

Previously an unknown `sk-ant-`-shaped key was forwarded to the LLM as a bring-your-own key. That path is gone. An `api_key` now resolves one of four ways:

- known client → swap in their stored Anthropic key
- known client with a `NULL` stored key → 500 (server misconfig, reported to Sentry)
- unknown key → 401, or 503 if the client store is unreachable (can't verify, so don't guess)
- no `api_key` → global `ANTHROPIC_API_KEY`, else 401

Removed the `forward` `KeyResolution` tag and the shape check that went with it. Also trimmed a couple of log lines that were either routine or not actionable: the internal-token mismatch no longer reports to Sentry (it's a config/forgery signal, not a bug), and a max-staleness cache eviction no longer warns (happens whenever the DB lags). Tidied the comments in `instance-auth.ts` while I was in there.

## AI Usage

- [x] Yes, I have used AI
- [ ] No, I have not used AI

### DIFF
--- a/.changeset/moody-coats-punch.md
+++ b/.changeset/moody-coats-punch.md
@@ -1,0 +1,5 @@
+---
+"apollo": patch
+---
+
+Remove anthropic token bypass for http auth
--- a/.changeset/moody-coats-punch.md
+++ b/.changeset/moody-coats-punch.md
@@ -1,5 +1,0 @@
----
-"apollo": patch
----
-
-Remove anthropic token bypass for http auth
--- a/.changeset/shy-dogs-switch.md
+++ b/.changeset/shy-dogs-switch.md
@@ -1,5 +1,0 @@
----
-"apollo": major
----
-
-Force authorisation on all API keys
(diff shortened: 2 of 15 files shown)


josephjclark:
QA notes:
- [x] echo service is rejected if no API key is passed
- [x] If ANTHROPIC_API_KEY is set, echo is allowed without a key
- [x] With ANTHROPIC_API_KEY unset, lightning requests without an api key will be rejected
- [x] With ANTHROPIC_API_KEY unset, lightning with a valid key will be rejected
- [x] With ANTHROPIC_API_KEY unset and a new API key cfreated, lightning requests are accepted

josephjclark (approved):

PR #565, opened 2026-06-29
Acceptance tests: Write evals to file and use experiment names

hanna-paasivirta:
## Short Description

Acceptance tests now save judge results to a file and let you label runs with -E so you can keep results from different runs side by side.

## Implementation Details

- This adds a third output file for acceptance test runs. It saves the judge verdicts to a .judges.txt file in tmp/, next to the YAML and response text. 
- You can also pass `--experiment=<name>` or `-E <name>` to tag the output files so different runs don't overwrite each other.

## AI Usage

know!):

- [x] Yes, I have not used AI
- [ ] No, I have not used AI


josephjclark:
Nice!

PR #573, opened 2026-07-08
Global chat: Add answer streaming to planner

### LINKED ISSUE #572: Global chat: Fix planner streaming
The planner does not actually stream its answer. Fix the ordering of sending YAML and streaming chunk event.

hanna-paasivirta:
## Short Description

The planner buffered all model text and flushed it only after the response finished. Users saw nothing while the planner worked. Text the model wrote before tool calls ("I'll build this for you...") was silently discarded. Now we stream the planner's narration text live and separate its progress messages into two event types: transient "...ing" spinners (thinking events) and a new custom `status` event for completed actions ("Edited workflow structure"), which also persist in a new `response_segments` field. Workflow YAML now streams the moment it changes instead of waiting for the next text.

Fixes #572 

Matching Lightning PR: https://github.com/OpenFn/lightning/pull/4969

## Implementation Details

Previously the planner buffered all text and sent it at the end, and every status was a thinking event with nothing to distinguish an in-progress spinner from a completed action. Lightning had no reliable way to know which statuses to persist, and the transcript could not be re-rendered after a page reload.

- Text deltas now stream as they generate, including the narration the model writes before tool calls.
- Each tool call sends a spinner ("Reviewing the workflow...") as a thinking event, then a past-tense settled line as a new custom `status` SSE event once the action completes. The settled verb reflects the actual outcome: a workflow-agent call that returned no YAML settles to "Analyzed the workflow" rather than claiming an edit, and job-code lines only count steps that were actually stitched in. Apollo owns this classification so Lightning just maps event type to visual treatment (spinners replace each other and vanish; status events persist).
- A new `response_segments` output field records the durable transcript (text woven with settled status lines, spinners excluded) so the frontend can re-render the same view on reload. The `status` event payload matches the segment shape, so one render path covers both.
- The YAML `changes` event is now sent inside the update itself, wherever `current_yaml` changes (workflow edit or job-code stitch), instead of being checked before every text delta. A newly added step reaches the client before its code is written, and the per-token change-detection machinery (`_sent_yaml`, pending-send checks) is deleted.
- Planner history now returns plain string content on every route, matching the direct routes.

## AI Usage

know!):

- [x] Yes, I have not used AI
- [ ] No, I have not used AI

### DIFF
--- a/.changeset/short-months-look.md
+++ b/.changeset/short-months-look.md
@@ -1,0 +1,6 @@
+---
+"apollo": minor
+---
+
+global chat: add answer streaming to the planner, breaking up responses into
+chunks which can be rendered earlier
--- a/services/global_chat/PAYLOAD_SPEC.md
+++ b/services/global_chat/PAYLOAD_SPEC.md
@@ -84,7 +84,12 @@ This document defines the input and output payload structure for the Global Agen
 
 ```json
 {
-  "response": "string",                  // Main text response
+  "response": "string",                  // Main text response (final answer)
+
+  "response_segments": [                 // Full transcript of the turn, in stream order
+    { "type": "status", "content": "Reviewing the workflow..." },
+    { "type": "text",   "content": "..." }
+  ],
 
   "attachments": [                       // Artifacts produced this turn
     {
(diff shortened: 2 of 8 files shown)


josephjclark:
Fantastic improvement. Even on lightning main, without the UI improvements, streaming is way more responsive and the UX is hugely improved

josephjclark (approved):

Issue #586, opened 2026-07-13
Add Langfuse metadata to show workflow changes


hanna-paasivirta:
Add Langfuse metadata to show workflow changes to distinguish what parts of the workflow were written by the assistant and what parts were already present. This is necessary to evaluate the generated workflow and code quality accurately.

Either add the diffs, or the full original workflow in the output payload metadata.

PR #587, opened 2026-07-13
Add langfuse original workflow

hanna-paasivirta:
## Short Description

Adds a `generation_diff` to the Langfuse trace metadata of `job_chat`, `workflow_chat` and `global_chat`, showing exactly what the model wrote in that turn versus what was already in the user's job code or workflow YAML.

Fixes #586 

## Implementation Details

**Problem.** Our Langfuse traces are evaluated for job code and workflow quality, but a trace only shows the final artifact. There is no way to tell which parts were generated in that turn and which were already there.

**Approach.** Each service now computes a unified diff between the original artifact from the request (job code from `context.expression`, workflow YAML from `existing_yaml` / `workflow_yaml`) and the generated one, and attaches it to the trace as metadata (`generation_diff` plus a small `diff_stats` count). A diff is the minimal representation that still shows everything needed: added lines are the model's work, context and removed lines are pre-existing. The original artifact itself is not stored, since it can always be reconstructed from the final output plus the diff.

**Only touches Langfuse metadata and not the response payload.** This is purely for observability, so it goes only on the trace via `update_current_span(metadata=...)`. Putting it in the response `meta` would bloat every payload back to Lightning with data Lightning does not use, and would change the public response shape. The Langfuse SDK flushes in the background, and the diff is computed after generation completes, so request latency is unaffected. Everything is gated on `metrics_opt_in`, so opted-out requests do no extra work.

## AI Usage

know!):

- [x] Yes, I have not used AI
- [ ] No, I have not used AI

### DIFF
--- a/.changeset/good-months-tease.md
+++ b/.changeset/good-months-tease.md
@@ -1,0 +1,5 @@
+---
+"apollo": patch
+---
+
+add generation diff metadata for langfuse
--- a/services/global_chat/global_chat.py
+++ b/services/global_chat/global_chat.py
@@ -14,7 +14,7 @@ sys.path.append(str(Path(__file__).parent.parent))
 
 from langfuse import observe, propagate_attributes, get_client as get_langfuse_client
 from util import ApolloError, create_logger, APOLLO_VERSION
-from langfuse_util import should_track, build_tags
+from langfuse_util import should_track, build_tags, build_generation_diff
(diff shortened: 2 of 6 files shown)


josephjclark (approved):

josephjclark:
Ah @hanna-paasivirta this now has merge conflicts. Please resolve and merge and then we can prep the release branch :)

PR #611, opened 2026-08-03
Create DocsiteSearch embeddings lazily instead of at import time

elias-ba:
## Short Description

Create `DocsiteSearch`'s default OpenAI embeddings client lazily so importing the module no longer crashes local setups that have no `OPENAI_API_KEY`.

## Implementation Details

`DocsiteSearch.__init__` used `embeddings=OpenAIEmbeddings()` as a default argument. Default arguments are evaluated at import time, and `OpenAIEmbeddings()` raises without `OPENAI_API_KEY` — and global chat imports this module through its tool definitions, so every global chat request crashed locally even when docsite search was never used.

The default is now constructed inside `__init__` (`embeddings=None` → create when needed). Behavior for real searches is unchanged; importing the module is side-effect free.

Verified: importing `search_docsite` without `OPENAI_API_KEY` succeeds on this branch and raises `openai.OpenAIError` on main. Existing `search_docsite` tests pass (9/9). Changeset included.

Found while testing OpenFn/lightning#4969 against Apollo locally.

## AI Usage

know!):

- [x] Yes, I have used AI (Claude Code)
- [ ] No, I have not used AI

### DIFF
--- a/.changeset/lazy-docsite-embeddings.md
+++ b/.changeset/lazy-docsite-embeddings.md
@@ -1,0 +1,7 @@
+---
+"apollo": patch
+---
+
+search_docsite: create the default OpenAIEmbeddings client lazily so
+importing the module (e.g. via global chat's tool imports) no longer
+crashes without OPENAI_API_KEY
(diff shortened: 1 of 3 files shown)


hanna-paasivirta (approved):
Thank you Elias, this is great. I've added a little unit test for it as well.

PR #619, opened 2026-08-11
Planner prompt change: cross-step data contracts

### LINKED ISSUE #569: Global chat: Workflow coherence
We have no system in place to check for workflow coherence. This is a problem particularly when generating a workflow with many steps from scratch.

Possible solutions:
- Adjust planner prompt to improve subagent instructions (minimal solution, probably not enough: tell it to specify the previous and following steps in more detail
- Adjust planner prompt to tell it to read through the workflow and then call subagents with required corrections
- Add a new subagent task: proofread all job code together; e.g. adjust job code agent so it can also receive several steps
- Track state between steps in the internal workflow representation

hanna-paasivirta:
## Short Description

Adds golden tests of representative workflows, which surface a key quality issue in the global assistant: cross-step drift in job code generated in parallel. A prompt change allocates the responsibility of defining key and field names to the planner, before calling the job code agent instances. Another prompt tweak reinforces the planner/subagent boundary to prevent the planner from defining code structure except when specified by the user. 

Fixes #569 

###  Problem

When the planner builds a workflow, each step's code is written by an isolated job-code agent call that cannot see any other step. Steps therefore disagreed on how data passes between them, producing workflows that look right but fail at runtime. In the baseline 8-step fan-out run, the validate step wrote the patient to `state.event`, a later step read `state.validatedEvent`, and the FHIR step read `state.data` — every downstream field `undefined`. In the baseline Kobo run the steps shared a key but not its fields: validation stored `sex` and `dateOfBirth`, the update step read `gender` and `birthdate`.

###  Change

One block of the planner system prompt (`prompts.yaml`, "Tool Call Ordering"). The baseline had the planner describe shared data by a label only and forbade mentioning state. Now the planner puts an identical contract line in every job-code message: the exact state key (e.g. `state.patient`), the exact field names inside it (from the user's message when given, otherwise chosen once), and ownership — one step produces it, downstream steps consume it verbatim and translate into their own target system's format internally. How the code produces or consumes it stays the job agent's call. (An intermediate version fixing only the key still left field drift; only key + fields closed both.) This is the lightweight form of the mapping spec human OpenFn implementers write before coding — at this scale it is 5–30 field names.

###  Results

We generated the same six workflows (three scenarios, each from a vague and a verbose prompt) under the baseline prompt and the final prompt, and read each generated YAML checking only the issue this change targets: do the steps agree on where shared data lives? **Coherent** means every `state.X.y` a step reads was actually written by an upstream step under exactly that key and field name, so the data arrives at runtime. The two failure modes are **key drift** (steps use different state keys for the same data — the object never arrives) and **field drift** (same key, different field names inside it — the object arrives with empty fields). This is judged by reading the code, not by test pass/fail.

| Scenario (steps) | Baseline | Final |
|---|---|---|
| REST-to-REST vague (2) | coherent | coherent |
| REST-to-REST verbose (2) | coherent | coherent |
| Kobo→OpenMRS vague (2) | coherent | coherent |
| Kobo→OpenMRS verbose (5) | **field drift** — update reads `registration.gender`/`birthdate`, validate stored `sex`/`dateOfBirth`; error handler reads `submissionId` where the source has `_id` | coherent* |
| Fan-out vague (3–4) | coherent | coherent |
| Fan-out verbose (8) | **key + field drift** — `state.event` vs `state.validatedEvent` vs `state.data` across steps, plus `gender` vs `sex`; reproduced in two independent runs | coherent — all 8 steps read `state.patient` with the exact contract fields; `gender`/`birthDate` appear only inside target payloads (correct translation) |

\* One slip on an auxiliary value outside the contract (lookup stored a UUID string, update read `.uuid` off it); the contract object itself is clean.

Drift only ever appeared in the two verbose multi-system workflows — the ones with the most independently generated steps — and the final prompt eliminated it in both. The generated YAML for every cell is in `services/global_chat/tests/acceptance/gold_workflows/tmp/` (`__baseline` / `__fields-v2` labels).

###  Second change: requirements, not implementation

We found the planner occasionally prescribing code structure to subagents (e.g. telling one to put "placeholder constants at the top", which the job agent obeyed, producing invalid top-level declarations). The old guardrail only banned suggesting function names — an enumeration that misses everything not listed. It is replaced with the boundary itself: messages state what the step must do, must never do, and what it reads/writes per the contract — never what the code should look like; anything the user explicitly specified is a requirement and is relayed verbatim. Verified on two runs: the Kobo verbose scenario no longer gets structure prescriptions in planner messages, and the existing function-relay acceptance test (user names `each()`, `fields()`, `tracker.import` with `CREATE_AND_UPDATE`) still relays everything verbatim and now passes all criteria.

###  Future

For complex workflows a future version could generate a real, user-reviewable mapping artifact (including code-list/terminology mappings) or let job agents read upstream code directly. Not needed for workflows at this scale.

## AI Usage

know!):

- [x] Yes, I have not used AI
- [ ] No, I have not used AI

### DIFF
--- a/services/global_chat/tests/acceptance/gold_workflows/test_patient_fanout_verbose.md
+++ b/services/global_chat/tests/acceptance/gold_workflows/test_patient_fanout_verbose.md
@@ -1,0 +1,230 @@
+---
+id: global-chat.gold.patient-fanout-verbose
+service: global_chat
+judges: [general, openfn_workflow_expert, openfn_code_quality]
+---
+
+# notes
+
+Verbose phrasing of the large fan-out pattern: a webhook-delivered
+registration event is validated, deduplicated against OpenMRS, created or
+updated there (with a generated identifier on the create path), then fanned
+out to DHIS2 and a FHIR server, with error paths for invalid input and
+downstream failures. The user spells everything out, so the response should
+honor the explicit instructions. The ones that carry the test:
+
+- "Do not add a fetch step" is explicit: the event arrives in `state.data`
+  via webhook, so an added fetch from the source system is a clear failure.
+- The idgen requirement is called out hard in the prompt: the OpenMRS ID has a
+  check digit and "must be generated, not made up". Code that fabricates an ID
+  (random string, timestamp, national ID reused as the OpenMRS ID) instead of
+  requesting one from the idgen source is a clear failure. On the create path
+  the generated ID is the preferred identifier and the national ID is a
+  second, non-preferred identifier.
+- Validation gates the writes: missing required fields (given name, family
+  name, sex 'M'/'F', date of birth, national ID) fail the run to an
+  invalid-input handler, and invalid events must never reach any write step.
+- Ordering and convergence: DHIS2 and FHIR run only after the OpenMRS write
+  succeeded, on either the create or the update path. Cross-standard mapping
+  matters at the FHIR step: sex 'M'/'F' maps to gender 'male'/'female'.
+- PII discipline is explicit: only the submission id is ever logged, never
+  names, dates of birth or national IDs, including in both error handlers.
+
+Because OpenFn workflows are trees (a step has one parent), candidates may
+express create-vs-update as in-job conditional logic on a single path, or as
+branch edges with the downstream fan-out duplicated under each branch. Either
+is acceptable as long as the paths are mutually exclusive and the fan-out runs
+after either one. Similarly the downstream-error handling may be one handler
+per failing step or any equivalent arrangement.
+
+The following workflow is a NON-BINDING reference showing one acceptable
+shape. Do not require the candidate to match it: job names, adaptor versions,
+exact function signatures (especially the idgen call and the DHIS2/FHIR
+payload details), and how branches and error routes are wired may all differ.
+Placeholder UUIDs are expected since the prompt says identifier type,
+location, org unit, program and attribute UUIDs are configured separately. Use
+it only to sanity-check that the candidate is a plausible, coherent solution;
+we want to catch clear issues against the explicit instructions above, not
+enforce this exact structure. Pagination and idempotency concerns from the
+broader rubric are not relevant to this single-event pattern.
+
+```yaml
+name: patient-registration-fanout
+jobs:
+  Validate-event:
+    name: Validate registration event
+    adaptor: "@openfn/language-common@latest"
+    body: |
+      fn((state) => {
+        const s = state.data;
+        const required = ['given_name', 'family_name', 'sex', 'date_of_birth', 'national_id'];
+        const missing = required.filter((f) => !s?.[f]);
+        if (missing.length > 0 || !['M', 'F'].includes(s.sex)) {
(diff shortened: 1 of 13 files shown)


hanna-paasivirta:
@elias-ba I don't think this requires a super thorough review on your part, as this is essentially a prompt change that we'll continue to iterate on for a while.

services/global_chat/tests/acceptance/gold_workflows/test_patient_fanout_verbose.md
```diff
@@ -0,0 +1,230 @@
...
+
+      fn((state) => {
+        const s = state.submission;
+        const person = {
+          names: [{ givenName: s.given_name, familyName: s.family_name }],
+          gender: s.sex,
+          birthdate: s.date_of_birth,
+        };
+        if (state.existingPatient) {
+          return update(`person/${state.existingPatient.uuid}`, person)(state);
```
elias-ba: Question on this bit. The create and update logic calls the operations inside `fn`, which gives `update(...)(state)` and `create(...)(next)`. Is that the pattern the code quality judge is told to flag? And since the reference workflow gets passed to the judge as test notes, could it pull the judge the wrong way on this test? Same shape in `test_patient_fanout_vague.md` around line 73.

Would the branch edge split you used in the Kobo reference work here too, or is there a reason it doesn't fit the fan-out?

services/global_chat/tests/acceptance/gold_workflows/test_rest_to_rest_daily_vague.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+# quality_criteria
+
+- A complete workflow is produced (assumptions or placeholders are fine), rather than only clarifying questions, since the request names both endpoints and a schedule.
+- The trigger is a daily cron (a morning schedule such as `0 6 * * *`, or any once-a-day expression), not a webhook.
+- A step fetches the user list from `https://jsonplaceholder.typicode.com/users` with an HTTP get.
+- Each user is reduced to some smaller object. Any sensible choice of fields is fine; do not require specific field names since the user only said "simplify".
+- Each transformed record is POSTed to `https://jsonplaceholder.typicode.com/posts`.
+- Job code uses real http adaptor functions with plausible arguments, not invented helpers, and each step returns state.
```
elias-ba: Trying to square this with the judge changes in the same PR. This criterion asks the judge to catch invented functions, and the judge files now say not to flag unless the docs contradict. Since they also say test criteria win on a conflict, does this put the guessing back whenever the docs lookup comes back empty? Or is there something about how docs get attached in these runs that I'm missing?

Same wording in `test_patient_fanout_vague.md`, `test_kobo_to_openmrs_vague.md` and `test_rest_to_rest_daily_verbose.md`. If it is a real conflict, would rewording to something checkable without docs make sense?

elias-ba (changes requested):
This looks good to me. The diagnosis matches what the code actually does: batched job code calls all get the same pre-batch YAML, so those steps really are blind to each other. Two small things in the test files below.

elias-ba (approved):

PR #623, opened 2026-08-12
Clarify docsite index setup in docs and fix dependencies

hanna-paasivirta:
## Short Description

Documents that each environment populates its own Pinecone docsite index, since there is no shared one to point at. Also fixes the Pinecone env var name in `.env.example` and the `bun py` commands in the RAG service READMEs.

## AI Usage

Please disclose whether you've used AI in this work (it's cool, we just want to
know!):

- [x] Yes, I have not used AI
- [ ] No, I have not used AI

You can read more details in our
[Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)

PR #627, opened 2026-08-14
embed_docsite: download docsite via git clone instead of GitHub contents API

### LINKED ISSUE #182: embed_docsite: use git clone to download source files from repo

IZO-Ong:
## Short Description

Replaces `embed_docsite`'s recursive GitHub Contents API walk with a shallow, blobless and sparse `git clone` of `OpenFn/docs`.

Fixes #182 

## Implementation Details

The old `github_utils.py` fetched `general_docs`/`adaptor_docs` one directory at a time via `api.github.com/.../contents/{path}` (~22 requests for OpenFn/docs, against a 60/hour unauthenticated limit), then downloaded each file individually. This PR ships two commits 4315b60 (for naive `git clone --depth 1`) and 4508e30 (for a more optimized git clone using `--filter=blob:none --sparse`, checked out to only `docs/` and `adaptors/`, thus avoiding downloading static assets and images).

`adaptor_functions` is untouched.

### Measurements

| Variant | Cold | Warm | Disk | Docs | GitHub REST calls |
|---|---|---|---|---|---|
| main (current) | 102.9s | n/a (no cache) | 0 | 191 | ~44/run |
| Trees + ETag cache (PR #618) | 92.1s | 1.02s | ~3MB + manifest | 191 | 2/run |
| Naive `git clone --depth 1` (commit 1) | 11.94s | 2.23s | 274MB | 191 | 0 |
| Blobless sparse clone (commit 2) | 4.01s | 1.27s | 2.9MB | 191 | 0 |

The sparse clone's corpus was verified byte-identical to the naive clone's by hashing the full `get_docs` output (sha256 of sorted-key JSON).

### Next steps
If the optimized, blobless / sparse `git clone` is ok, we can change the command for `dockerfile` on line 17 to use the optimized version instead.

## AI Usage

know!):

- [X] Yes, I have used AI
- [ ] No, I have not used AI


josephjclark:
Hey @IZO-Ong I've started to run some tests all this and it already feels a lot faster than prod. I got an error at the end but I think that's just my local setup, I'm gonna explore it now.

I'll try and get full testing and review done today. I might have to ask you to rebase on a different PR which adds an important fix that's related.

services/embed_docsite/docs_repo.py
```diff
@@ -0,0 +1,127 @@
+"""A shallow, blobless, sparse git checkout of OpenFn/docs, kept on disk.
```
josephjclark: shouldn't all this stuff just live in `github_utils`? Or the rest of `github_utils` should be copied in here? Both  files serve the same purpose

services/embed_docsite/docs_repo.py
```diff
@@ -0,0 +1,127 @@
...
+DOCS_TYPE_PREFIXES = {"general_docs": "docs", "adaptor_docs": "adaptors"}
+
+GIT_TIMEOUT_SECONDS = 300
+
+# Memoized per process: a full run reads two docs_types, and they should
+# cost one sync, not two.
+_synced_sha = None
+
+
+def _run_git(args, cwd=None):
```
josephjclark: I'm not a python guy but I'm really not a fan of the private underscored convention. it looks ugly and doesn't really mean anything in an internal module like this. Why pretend its private? It can still be used  externally by someone who needs it.

@hanna-paasivirta I don't think we use this convention very much across apollo?

services/embed_docsite/docs_repo.py
```diff
@@ -0,0 +1,127 @@
...
+    """Subprocess seam."""
```
josephjclark: What does this comment mean? What is its value?

I know it's AI generated (not least because claude's obsessed with seams lately) but it still affects the reader. And it's just a pointless, unhelpful comment with no value that interrupts the reading of the function.

services/embed_docsite/docs_repo.py
```diff
@@ -0,0 +1,127 @@
...
+        text=True,
+        timeout=GIT_TIMEOUT_SECONDS,
+    )
+
+
+def _has_checkout():
+    return (CLONE_DIR / ".git").exists()
+
+
+def _clone():
```
josephjclark: maybe we can rename the function wipe and clone? Or take the wipe bit as an argument. Because it's quite a big side effect... I wouldn't expect `clone` to be distructive

services/embed_docsite/docs_repo.py
```diff
@@ -0,0 +1,127 @@
...
+CLONE_DIR = Path(__file__).parent / "docsite_cache"
+
+# Which top-level directory in OpenFn/docs backs each docs_type.
+DOCS_TYPE_PREFIXES = {"general_docs": "docs", "adaptor_docs": "adaptors"}
+
+GIT_TIMEOUT_SECONDS = 300
+
+# Memoized per process: a full run reads two docs_types, and they should
+# cost one sync, not two.
+_synced_sha = None
```
josephjclark: Not a big fan of this variable name - can we think of a non-underscored name that would be self explanatory without  comment?

BTW each call to `embed_docsite` will run in a different python process right now, so this value will be lost. That might change in future so we'll keep the logic.

services/embed_docsite/docs_repo.py
```diff
@@ -0,0 +1,127 @@
...
+        ["git", *args],
+        cwd=cwd,
+        check=True,
+        capture_output=True,
+        text=True,
+        timeout=GIT_TIMEOUT_SECONDS,
+    )
+
+
+def _has_checkout():
```
josephjclark: I think this only has one caller. having this as its own function makes the code harder to read, not easier!

services/embed_docsite/docs_repo.py
```diff
@@ -0,0 +1,127 @@
...
+    _run_git(["reset", "--hard", "FETCH_HEAD"], cwd=CLONE_DIR)
+    logger.info("Docs checkout updated")
+
+
+def _head_sha():
+    result = _run_git(["rev-parse", "HEAD"], cwd=CLONE_DIR)
+    return result.stdout.strip()
+
+
+def sync_docs_repo():
```
josephjclark: I would rather that the public API was at the top of the file, so that readers find the most relevant logic and APIs first without having to hunt for it

josephjclark (changes requested):
Looking great @IZO-Ong thank you for this

I've left some feedback about structure and style.

We probably need to rebase this on top of  https://github.com/OpenFn/apollo/pull/623 because I think they overlap a bit.

I think maybe  first you should refactor here per my comments, then we should merge this and 623 into a merge branch, and then into main.

CC @hanna-paasivirta

services/embed_docsite/docs_repo.py
hanna-paasivirta: I personally find it helpful for understanding the code sometimes, but perhaps not necessary here

services/embed_docsite/docs_repo.py
```diff
@@ -0,0 +1,127 @@
...
+    Calls `sync_docs_repo` first.
+
+    :return: [{"name": <basename>, "docs": <text>}], sorted by path
+    """
+    if docs_type not in DOCS_TYPE_PREFIXES:
+        raise ApolloError(400, f"Unknown docs_type '{docs_type}'", type="BAD_REQUEST")
+
+    prefix_dir = CLONE_DIR / DOCS_TYPE_PREFIXES[docs_type]
+    paths = sorted(prefix_dir.rglob("*.md"))
+    return [{"name": path.name, "docs": path.read_text(encoding="utf-8")} for path in paths]
```
hanna-paasivirta: The old download_main_docs caught per-file errors, logged, and skipped the file if there was an issue. Should we wrap the read in a per-file try/except (log+skip) to keep that behaviour?

services/embed_docsite/docs_repo.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Calls `sync_docs_repo` first.
+
+    :return: [{"name": <basename>, "docs": <text>}], sorted by path
+    """
+    if docs_type not in DOCS_TYPE_PREFIXES:
+        raise ApolloError(400, f"Unknown docs_type '{docs_type}'", type="BAD_REQUEST")
+
+    prefix_dir = CLONE_DIR / DOCS_TYPE_PREFIXES[docs_type]
+    paths = sorted(prefix_dir.rglob("*.md"))
```
hanna-paasivirta: Would this fail silently and leave in corrupted docs? We could raise an error here like

`raise ApolloError(500, f"No markdown files found under {prefix_dir} checkout may be broken", type="INTERNAL_ERROR")`

IZO-Ong:
Hi all, I've updated the branch based on your comments (thank you!). Please feel free to re-review the changes so that we can merge / rebase into PR #623 if needed.

josephjclark:
grr didn't mean to close this

josephjclark (approved):

PR #631, opened 2026-08-15
Say what actually went wrong when a service fails

elias-ba:
## Short Description

Makes a failing service report a real code, type and message, instead of arriving at the client as "Unknown error".

Stacked on #630 → #629. Review those first; this branch contains them.

## Implementation Details

Three ways a failure lost its own diagnosis.

**A bare exit code.** `bridge.ts` rejected with `code`. A number is not an `Error`, so the handler downstream fell through to `error instanceof Error ? error.message : "Unknown error"` and sent the client exactly that - no code, no type, and the exit code discarded. It also rejected without returning, so a non-zero exit went on to resolve as well.

**A failed spawn was only logged**, so nothing settled the promise: poetry missing from PATH looked like a service that never answered, and the request hung until something upstream gave up.

**Empty output resolved as success**, sent as `event: complete` with a null payload, which Lightning reports as a stream that ended without a response.

Two of those empty output files came from `entry.py`: three error paths returned before the output file was written, and two referenced an unbound exception variable, so a missing input file raised `NameError` rather than reporting the missing file. `ruff` flags both - two `F821 Undefined name 'e'` findings on `main`, gone here. Every path out of `run_service` now goes through one write, which is what makes an empty file mean the run died rather than that it failed politely.

**Shape.** `ApolloThrowable` extends `Error` while implementing the existing `ApolloError` interface, so it can be thrown and still carries the envelope used everywhere else. `toJSON` is required: `JSON.stringify` on an `Error` is `{}`, so without it the synchronous route answers with the right status and an empty body. There's a test on that.

Three factories: `SUBPROCESS_FAILED` (500, carries the exit code), `SUBPROCESS_SPAWN_FAILED` (500, never started), `EMPTY_RESULT` (502 - it ran and exited cleanly, what came back was unusable).

The synchronous route also gains error handling, which it had none of.

`bun test platform/test`: 101 pass, 0 fail. `ruff check services/entry.py`: 26 → 24 findings, nothing new.

## AI Usage

know!):

- [x] Yes, I have not used AI
- [ ] No, I have not used AI


hanna-paasivirta:
Thank you Elias, this is great. I've added a couple of things, could you confirm if these look ok:

1. JSON.parse(text) can still hang the request. If the output file holds broken JSON (e.g. the process died mid-write), the parse error is thrown inside the async close callback, nothing catches it. Now caught and rejected as MALFORMED_RESULT (502)

2. A killed process was misreported. code is null when the process dies by signal, so OOM or a deploy’s SIGTERM skipped if (code) and fell through to EMPTY_RESULT — “finished without producing a result”, which sends whoever’s debugging to the wrong place. The handler already receives signal; it’s now checked first and rejected as SUBPROCESS_KILLED with the signal name in details.

PR #632, opened 2026-08-15
Stop a run when the client goes away

elias-ba:
## Short Description

Kills a service run when the client disconnects, so an abandoned request stops calling the model.

Stacked on #631 → #630 → #629. Review those first; this branch contains them.

## Implementation Details

Nothing told us the caller had left. The python child kept running, kept calling Anthropic, and wrote its answer into a socket with nobody on the other end — potentially another ten model calls after the user closed the panel, all billed.

The `ReadableStream` had a `start()` and no `cancel()`, and the child process handle never escaped the bridge's promise closure, so there was nothing to stop even if we had noticed.

The stream now signals an `AbortController` when cancelled, and the bridge kills the child on that signal. `poetry run` execs into python rather than forking it, so the pid we hold is the interpreter and a plain `SIGTERM` reaches it — no process-group handling needed. Python installs no `SIGTERM` handler, so it dies immediately, closing its socket to Anthropic, which aborts generation. A `SIGKILL` follows five seconds later only for a child wedged somewhere that never sees the first signal.

Two things worth being precise about:

- **The saving is on streaming calls only.** A non-streaming `messages.create` has already been submitted, and Anthropic bills the completion whether or not anyone is listening.
- **A cancelled run settles as `SUBPROCESS_CANCELLED` (499), not a 5xx.** Abandoning a request shouldn't read as something breaking or pollute the error rate.

We listen on the request's own abort signal as well as the stream's `cancel()`, since the latter depends on the runtime noticing the dropped connection. `abort()` is idempotent.

## Also in this branch

Review turned up several things in the same area. They are here rather than in their own PRs because they touch the same lines and several only make sense once cancellation exists:

- **Cancellation reached only the streaming route.** A plain `POST` and a websocket both left the model generating when the caller went away. Both pass a signal now; the socket gets its own controller and a `close` handler.
- **The run promise could fail to settle.** `new Promise(async ...)` only catches a *synchronous* throw from its executor, so a failing `Bun.write` left it pending for ever and the caller's stream open. Setup moved outside the promise, where `run()` being async is enough.
- **A child killed by a signal reports a null exit code**, so the OOM killer — the likeliest way a service dies without exiting — was reported as an empty result with the signal discarded. It gets its own `SUBPROCESS_KILLED` envelope.
- **Every stderr line was forwarded to the caller** with no filter, so an interpreter traceback carrying server paths went straight out. Stderr now follows the same rule as stdout: only what a service logged deliberately.
- **The health check calls `run()` too**, and was the one caller that never got a catch when `run()` started rejecting.
- **The heartbeat interval had a floor but no ceiling**, and `setInterval` turns a delay past 2^31-1 into "every tick" — so the value someone picks to mean "effectively never" would have flooded every open stream.
- **Websockets have their own idle timer** that `serve.idleTimeout` does not reach; it defaulted to 120s, so the route most likely to be waiting on a slow answer was still being dropped.
- **Temp payload files are 0600 and swept at startup.** Only the close handler removed them, so anything that stopped the process mid-run left one behind in a directory nothing sweeps.

## AI Usage

- [x] I have used Claude Code


hanna-paasivirta:
Looks great! I added two tweaks:
1. The websocket cancellation didn’t seem to fire. Elysia creates a brand-new ws wrapper object for every callback, so the object close() gets is never the object message() stored in the WeakMap — the lookup always missed and the python child kept running. The fix is to key on ws.data (the upgrade context, created once per connection) instead of ws. The existing bridge tests couldn’t catch this because they hand run() a signal directly and never touch the route wiring, so I added a test that opens a real socket, closes it, and checks the child dies. It fails on your original code, passes with the fix.

2. I exempted two services from the cancellation. Killing embed_docsite mid-run leaves a half-filled Pinecone namespace, and because search_docsite always reads the newest namespace, global chat would silently start searching partial docs. And load_adaptor_docs commits its DELETE before its INSERT, so a kill in the gap erases an adaptor’s docs entirely. Both now always run to completion (i.e. they keep the pre-PR behavior). Small NON_CANCELLABLE set in services.ts with a comment.

PR #635, opened 2026-08-16
Mask sensitive values on their way out of a service

### LINKED ISSUE #634: echo: stop reflecting the full payload
While reviewing the timeout work I noticed `echo` returns its input verbatim and logs it. Payloads that reach a service can carry values the server set rather than the caller, so reflecting one wholesale is not a good default.

#635 covers this:

- `echo` runs its payload through the `mask_secrets` helper that already existed for Langfuse traces, and that helper now covers every field the server may fill in rather than only the one it fills in today
- masking moved to the boundary rather than sitting in one service, so a service that reflects its payload cannot leak by forgetting: service loggers mask what they emit, and every result leaves through one masked exit
- a test drives that exit through an unmounted probe that reflects its payload and masks nothing itself, so it fails if the boundary stops working rather than merely proving `echo` behaves

Closing when that merges.

elias-ba:
## Short Description

A payload reaching a service carries values the server put there rather than the caller, and services could hand them back out. This masks them at the boundary instead of trusting each service to remember.

Stacked on #632 → #631 → #630 → #629. Review those first; this branch contains them.

## Implementation Details

Three routes led back to the caller:

- **`echo` returned its input verbatim**, which is what started this
- **any service logging its payload** reaches the caller too, because logger output is forwarded as SSE `log` events — and `vocab_mapper` logs its whole payload on the first line of `main`
- **the error envelope returns the exception text**, and nearly every service catches broadly and rewraps as `ApolloError(500, str(e))` with the request in scope

Rather than patch three call sites, masking moved to places a service cannot bypass:

- `create_logger` installs a filter on the log handler, so what a service uses to log makes no difference
- every result leaves through one masked exit in `entry.call`
- a `before_send` hook masks what Sentry receives — it scrubs frame locals by name, but not exception text, contexts or breadcrumbs

`mask_secrets` already existed for Langfuse traces, so it is reused rather than duplicated. Doing so showed it was narrower than what the server can fill in: it now covers the other providers' field names, normalised so `api-key`, `apiKey` and `X-Api-Key` all match, and has a depth bound so a deeply nested payload cannot turn into a `RecursionError`.

The value pattern needed care in both directions, because the same function passes over workflow YAML, job code and service results. A loose version ate `task-`, `risk-` and `disk-` prefixed names; a different loose version ate `sk-antelope-migration-plan`. It anchors on the prefixes providers actually use, or wants a run too long and unbroken to be a name, and `test_mask_secrets.py` pins both directions.

## Additional notes for the reviewer

**Five instance-auth tests changed, and the coverage they had is restored elsewhere.** They read the substituted value off `echo`'s response, which masking makes impossible — with everything coming back `[REDACTED]`, a swap that wrote the wrong value, or forwarded the caller's own, would look identical to a correct one. So `applyResolvedKey` is now its own exported function and asserted directly: the substitution is checked without a service having to hand a value back to prove it. Paired with the `InstanceAuth` rows that pin which resolution a given credential produces, that covers the same ground the round trip did. Verified by breaking it both ways — forwarding the caller's own credential fails two of the five new assertions, blanking the field rather than dropping it fails one.

**A websocket fix rides along.** Live testing found `close` was aborting nothing: Elysia builds a fresh wrapper object per event, so the object the close handler receives is never the one the message handler stored against. Keyed on the underlying socket now, with a test that fails without the fix.

Closes #634

## AI Usage

- [x] I have used Claude Code


hanna-paasivirta:
Looks great, thank you Elias! I added a tiny change to copy the logger registry before walking it `list(root.manager.loggerDict.values())` so that if a logger is created by another thread it won't crash in a really confusing way. Let me know if this is ok and I'll merge this one @elias-ba

elias-ba:
Thanks @hanna-paasivirta that sounds great, nice catch. Please feel free to merge PRs when you happy with them

platform/test/server.test.ts
```diff
@@ -544,8 +553,8 @@ describe("Instance authentication", () => {
   // that ws.data carries the lightningClient set during beforeHandle.
   it("swaps a known client's stored key on a WS upgrade via ?api_key=", async () => {
     const body = await wsRoundTrip(`?api_key=${encodeURIComponent(ALPHA)}`);
-    expect(body.api_key).toBe("sk-ant-stored-alpha");
-    expect(body.api_key).not.toBe(ALPHA);
+    expect(body.api_key).toBe("[REDACTED]");
```
josephjclark: These tests are kind of broken now no? I'd need to read up on what's happening here by my Spidey senses are tingling a bit

